Issue #16, opened 2020-03-10
Schema needs to be much more robust


robin-macpherson:
## Description

The current schema is very basic and only supports two tables and data types. This needs to be built out significantly with carefully crafted design and relationships.

This issue is exploratory by nature but should handle support for:
- Profile
- Location
- Topic
- Tag
- Like
- Comment
- Language

As well as improving the general design -- `Published` can be `PostStatus`, for example.

## Before Filing

- [x] I have checked for duplicate issues
- [x] I have filled out this template properly

PR #17, opened 2020-03-10
Improve Database Design & Schema

robin-macpherson:
## Description

**Issue:** #16 

This PR builds out our schema significantly to add support for a number of new models. It also seeks to improve the general database design by carefully crafting relationships and thinking about what should be its own table and what should be a scalar type.

## Diagram

**A diagram to help visualize how the `many-to-many` relationships are working between `Users` and `Languages`:**

Journaly Users To Languages M2M Relationship

*This is how it works with both the `LanguageToNative` and `LanguageToLearner` relationship behind the scenes, with an intermediary table to handle the `m2m`.*

## Sub-Tasks

- [x] Add new models
- [ ] generate new migrations
- [ ] Update `graphql.ts` as minimally as needed
- [ ] Flesh your Queries & Mutations a little
- [ ] Update `seed.ts` script


robin-macpherson:
@nlitwin still a work in progress, but sharing the PR with you in case you want some early insight into the direction of the improvements to our database schema 😄🚀 I drew a nice diagram to help explain/visualize the `m2m` relationships I'm setting up between `Users` and `Languages` in particular. Enjoy and looking forward to your early feedback!

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
-  CreatedAt  DateTime @default(now())
-  UpdatedAt  DateTime @updatedAt
+  CreatedAt  DateTime   @default(now())
+  UpdatedAt  DateTime   @updatedAt
   author     User
+  ReadCount  Int        @default(0)
+  ReadTime   Int
+  likes      Like[]
+  comments   Comment[]
+  tags       Tag[]
```
nlitwin: You mentioned tags when we chatted in person. I think it's a cool idea, but maybe a feature we'd want to add at a later date? I feel like topics would be plenty for now. Watcha think?

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
+  topic      Topic[]
+  language   Language
```
nlitwin: I just had a thought... what if people want to write posts in multiple languages?? Maybe something we could add support for later if people asked for it. I'm thinkin' we stick to simple and see how it goes/what users ask for

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
+  Name              String
+  Email             String        @unique
+  Password          String
+  ResetToken        String?
+  ResetTokenExpiry  Float?
+  userRole          UserRole      @default(FREE_USER)
+  CreatedAt         DateTime      @default(now())
+  UpdatedAt         DateTime      @updatedAt
+  Handle            String        @unique
+  Bio               String
```
nlitwin: Should this be optional?

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
+  SocialMedia       SocialMedia[]
```
nlitwin: Can this type of field be optional too? if it's the `[]`/many type? Or is it implicitly optional?

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
+  posts Post[]
+}
+
+model Tag {
+  Id    String @id @default(cuid())
+  Name  String
+  posts Post[]
+}
+
+model Like {
```
nlitwin:
```suggestion
model Love {
```

j/k 🤣

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
+  ResetTokenExpiry  Float?
+  userRole          UserRole      @default(FREE_USER)
+  CreatedAt         DateTime      @default(now())
+  UpdatedAt         DateTime      @updatedAt
+  Handle            String        @unique
+  Bio               String
+  SocialMedia       SocialMedia[]
+  interests         Topic[]
+  location          Location?
+  languagesNative   Language[]    @relation("LanguageToNative")
```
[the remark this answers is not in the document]
robin-macpherson: The way I'm doing it is more thought-out than that, but still totally valid discussion 😊 The convention I'm using currently is that fields that are a scalar type are PascalCased and types that are custom types are camelCased, which gives you an indication just looking at the field name whether there's a relationship or not.

I wouldn't want to do `snake_case` here because this is going to end up as generated TypeScript/JavaScript code and we'll be potentially using those operations in our resolvers, etc.

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
-  Password         String
-  ResetToken       String?
-  ResetTokenExpiry Float?
-  status           Status[]
-  posts            Post[]
-  CreatedAt        DateTime @default(now())
-  UpdatedAt        DateTime @updatedAt
+model SocialMedia {
+  Id       String              @id @default(cuid())
+  Platform SocialMediaPlatform
```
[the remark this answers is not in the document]
robin-macpherson: Yeah I haven't finished building some of these out just yet so it might be a little confusing, just wanted to post so that you can see what I'm working on and the progress 😄 The `SocialMedia` model is for storing information about a particular platform. Users will have the option to add their social media info if they would like to, so they might add their LinkedIn and Instagram for example; or perhaps they have a YouTube channel. This will end up being a clickable icon that leads to the corresponding page/link. 

We're specifically supporting Facebook, YouTube, Instagram, and LinkedIn. We may have it in the code so that when they go to write in their link, it pre-populates most of the link so they only have to add their unique portion. Take a look at the model now with the Url field 😊🔗

backend/prisma/schema.prisma
```diff
@@ -7,31 +7,110 @@ generator prisma_client {
...
 }
 
 model Post {
-  Id         String   @id @default(cuid())
+  Id         String     @id @default(cuid())
   Title      String
   Body       String
-  Published  Boolean
+  status     PostStatus @default(DRAFT)
   Image      String?
```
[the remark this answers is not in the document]
robin-macpherson: Yep yep, I remember this was a good suggestion you'd had previously back when we were doing the previous iteration. I didn't add it yet because I need to think a little about the best way to construct it, but it's on the list for this PR for sure! Will update you when that's ready to take another look at 🏎💨✨

backend/prisma/schema.prisma
robin-macpherson: Sounds good, took it out for now 👍🏼

backend/prisma/schema.prisma
robin-macpherson: Yeah I already thought about this and I actually think that it is a constraint I would like to keep for the foreseeable future. I don't think we need to support people writing a post in multiple languages -- it could add so many complications.

One of the strengths of Journaly architecturally is how simple and well-structured the data is going to be, which will be so helpful for building powerful search and matching capabilities and incorporating ML, for example. Posts are organized by languages and topics. Imagine how complex it could become both for users and for our tooling if suddenly posts can be in multiple languages.

For example, our future amazing tools for running realtime text analysis as a user types to help them improve their writing in French. Or users who speak English and are learning German and so have their personal feed all nicely set up to only show them those posts, but suddenly start encountering posts where they can only understand or provide feedback on one paragraph in the middle. I thought of a bunch more examples we can chat about.

Personally, I think it's totally fair for those users who want to write in multiple languages to simply have to write posts in each one 😊 We could provide cool features later that help them, such as a "translation" tool where they can write a post in their native language (or just paste some existing text they like) and then practice translating it into a language they're learning to get feedback, and they could do this for several languages if they wanted.

But anyway, those are my thoughts on this 🙂

backend/prisma/schema.prisma
robin-macpherson: Yep 👍🏼

backend/prisma/schema.prisma
robin-macpherson: List fields are implicitly optional since it will just start as an empty list.

backend/prisma/schema.prisma
robin-macpherson: Too late we're going with Love!

backend/prisma/schema.prisma
nlitwin: Cool, makes sense!

backend/prisma/schema.prisma
nlitwin: Those are definitely good points! If we for some reason did want to allow people to write in multiple languages without complicating the DB structure, we could actually have a language called "all" or "world" or "multiple". I've seen that strategy used for entities that are available everywhere instead of one country locale. Sounds good though to just keep it simple for now!

backend/prisma/schema.prisma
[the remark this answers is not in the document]
nlitwin: Definitely agree about `snake_case` - I thought maybe the fields would be converted automatically to `camelCase` for use in the frontend JS. But after seeing #28 and how it's used: ❌~snake_case~ ❌ 😆 

I think your idea of indicating scalar vs custom type with different `cases` is pretty clever 🦊 My only concerns are that it could be awkward using both cases when accessing values as properties, e.g. `post.Name` vs `user.languagesNative`, and conventionally, properties/fields are lowercase, with `PascalCase` implying that something is a Model or object instance. I also think it's actually a good thing to refer back to the schema file if there is ever a need for knowing if a field is scalar or custom. My vote would be to keep it simple and align with the style in the Prisma docs because it could be really hard to change later. Che ne pensi?

backend/prisma/schema.prisma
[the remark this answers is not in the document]
nlitwin: Ok that makes sense! So they could just add `nicklitwin` and then we'd construct the url to be `facebook.com/nicklitwin` 👍 

Is the Url field added to `SocialMedia`? I couldn't find it, but maybe it's just not pushed up yet?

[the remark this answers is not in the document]
robin-macpherson: Yep it's there, did you look at the files changed? It doesn't show updates to the comment preview. It's currently on line 32 🙂

backend/prisma/schema.prisma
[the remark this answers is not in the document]
robin-macpherson: Yep I'm fine with that! It was definitely more something I was trying out, but after using it for a while and seeing the experience on the client-side, too, I agree with your suggestion 😄

backend/prisma/schema.prisma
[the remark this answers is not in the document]
nlitwin: Doh! Not sure how I missed that. looks good to me :)

backend/prisma/schema.prisma
```diff
@@ -7,35 +7,238 @@ generator prisma_client {
...
+  user                User?    @relation(fields: [userId], references: [id])
+  userId              Int?
+  sessionId           String
+  pageName            String
+  timestamp           DateTime @default(now())
+  utmSource           String?
+  // TODO (robin-macpherson): find user agent string parsing library and store each part separately
+  user_agent_string   String
+  user_agent_language String
+}
```
robin-macpherson, line 220: @nlitwin going to extract this `PageView` model out of here in a coming PR because all of our event data will be getting stored separately via a pipeline we'll be setting up specifically for that. But this shows the pattern of having a model per major event type for now 🙂

backend/src/graphql.ts
```diff
@@ -27,10 +31,49 @@ schema.objectType({
...
+    })
+  },
+})
+
+// schema.objectType({
+//   name: 'LanguageLearned',
+//   definition(t) {
+//     t.
+//   }
+// })
```
robin-macpherson, line 75: @nlitwin just as a reminder I'm going to be cleaning all of this up and fixing it up piece by piece in a series of coming PRs that will each be specific and contained 😊

robin-macpherson:
Heyyyy @nlitwin ! So this is finally ready. Just as a recap from our conversation, the main focus of this PR is now the design of the database (excitedddd!!!). Woot! 💪🏼🙌🏼

backend/prisma/migrations/20200408071119-init/schema.prisma
```diff
@@ -0,0 +1,118 @@
+datasource db {
```
nlitwin, line 1: Just checking/curious, is this file and the `readme.md` file in `backend/prisma/migrations/20200408071119-init/README.md` needed or can we delete them?

robin-macpherson: Yep they are needed 🙂

robin-macpherson:
> 🥕🐰💨⚡️
> 
> Awesome!!! so excited to get this in. Great work with all the research that has gone into this!

Wooohoooooooo 🤩😄

backend/prisma/migrations/20200408071119-init/schema.prisma
nlitwin: @robin-macpherson ahh ok, just wanted to double check cause the column values were spelled differently, so I thought we might need to generate a new migration

PR #21, opened 2020-03-21
Upgrade `nexus-future` & move to relying on them for managing Prisma2 dependency

robin-macpherson:
## Description

The `nexus-future` package now ships with `prisma2` which greatly reduces our overhead and need to manage this rapidly changing dependency ourselves 🙌🏼 So I'm removing this from our `devDependencies` and moving to the version that ships with `nexus-future`.

Upgrade packages:
```
"nexus-future": "^0.12.0",
"nexus-plugin-prisma": "^0.5.1"
```

## Sub-Tasks

- [x] Perform upgrades
- [x] Test 🧪

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -23,13 +23,18 @@
         "@babel/highlight": "^7.8.3"
       }
     },
+    "@babel/helper-validator-identifier": {
+      "version": "7.9.0",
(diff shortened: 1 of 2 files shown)


robin-macpherson:
Merging this as it is safely tested and I need it to unblock me on my current branches 😊

PR #27, opened 2020-03-23
Add blog post

nlitwin:
## Description

Converting the Gatsby blog post into the default blog page for now. Using [date-fns](https://date-fns.org/docs) for formatting dates (a super popular, lightweight alternative to Moment.js)

## Screenshots

image

## Sub-Tasks

- [x] Create layout for blog posts
- [x] Convert `.md` blog post into HTML


frontend/config/index.ts
```diff
@@ -0,0 +1,4 @@
+import dayjs from 'dayjs'
+import advancedFormat from 'dayjs/plugin/advancedFormat'
```
nlitwin, line 2: [`advancedFormat`](https://github.com/iamkun/dayjs#plugin) allows formatting with `Do`. Other formats [here](https://day.js.org/docs/en/parse/string-format#list-of-all-available-parsing-tokens)

frontend/components/Layouts/BlogPageLayout.tsx
```diff
@@ -0,0 +1,88 @@
+import { format } from 'date-fns'
+import { getUTCDate } from '../../utils'
+
+type Props = {
+  date: string
+  title: string
+  image: string
+  children: React.ReactNode
+}
+
```
robin-macpherson, line 10: I'd like to use interfaces for defining simple objects and in particular props 😊 This is also in the style guide but I'd like to write more there shortly 😄

frontend/components/Layouts/BlogPageLayout.tsx
```diff
@@ -0,0 +1,88 @@
...
+const BlogPageLayout = (props: Props) => (
+  <div>
```
robin-macpherson, line 12: I'd also like to stick to being nice and explicit with our React components typings, so like I've been doing in all the recent PRs: `const BlogPageLayout: React.FC<Props> = () => {...}`

Also, and again a lot of this is just a matter of style and convention/continuity across the app, what do you think about destructuring the props in a case like this? 

`const BlogPageLayout: React.FC<Props> = ({ title, image, date, children }: Props) => {...}` ?

frontend/pages/blog/introducing-journaly.tsx
```diff
@@ -0,0 +1,295 @@
...
+        p {
+          margin-top: 16px;
+        }
+
+        p:first-child {
+          margin-top: 0;
+        }
+
+        a {
+          color: #4391c9;
```
robin-macpherson, line 282: We can use `brandBlue` from our `colors` util file 🔷

frontend/utils/date.ts
```diff
@@ -0,0 +1,14 @@
...
+
+  return new Date(
+    date.getUTCFullYear(),
+    date.getUTCMonth(),
+    date.getUTCDate(),
+    date.getUTCHours(),
+    date.getUTCMinutes(),
+    date.getUTCSeconds(),
+  )
+}
```
robin-macpherson, line 14: Nice!! Thanks for making this 😀🎈

frontend/utils/index.js
```diff
@@ -1,2 +1,3 @@
 export * from './breakpoints'
 export * from './colors'
+export * from './date'
```
robin-macpherson, line 3: 👌🏼😄

frontend/pages/blog/introducing-journaly.tsx
```diff
@@ -0,0 +1,295 @@
...
+        launch, and check out{' '}
+        <a href="https://youtube.com/user/TheLifeOfRob">my YouTube channel</a>{' '}
+        where I'll be posting updates along with my other language content.
+      </p>
+
+      <p>
+        Got questions? We'd love to talk. Email us at{' '}
+        <em>
+          <a href="mailto:hello@journaly.io">hello@journaly.io</a>
+        </em>
```
robin-macpherson, line 254: We can actually change this to `robin@journaly.com` now 🥳 and I'll set up a `hello` and other emails once we're about to launch (trying to limit all the things I'm paying for each month before we actually launch 😅). But let's change this to mine for now so that we don't risk forgetting to change it.

frontend/pages/blog/introducing-journaly.tsx
```diff
@@ -0,0 +1,295 @@
...
+      date="2019-09-25"
+      image={'../images/blog/post01-image.jpg'}
+    >
+      <p>
+        Hi there! I'm Robin. Long-time language enthusiast,{' '}
+        <a href="https://youtube.com/user/TheLifeOfRob">polyglot YouTuber</a>,
+        language teacher, and author of the book{' '}
+        <em>
+          <a href="https://www.kumabrand.com/shop">How To Maintain Languages</a>
+        </em>
```
robin-macpherson, line 17: HOPEFULLY when this launches, we'll be able to change this to `robinmacpherson.co/shop` 🤩 hehe

frontend/pages/blog/introducing-journaly.tsx
```diff
@@ -0,0 +1,295 @@
+import LandingPageLayout from '../../components/Layouts/LandingPageLayout'
+import BlogPageLayout from '../../components/Layouts/BlogPageLayout'
+
+const IntroducingJournalyBlogPost = () => (
```
robin-macpherson, line 4: `import { NextPage } from 'next'`

`const IntroducingJournalyBlogPost: NextPage = () => (..)`

frontend/components/Layouts/BlogPageLayout.tsx
nlitwin: Oh I think I may have done that before seeing the style guide. I read somewhere that types are better to use as a default than interfaces, but it seems like interfaces are the most popular way with React Props, so let's go with interfaces! 👍

frontend/components/Layouts/BlogPageLayout.tsx
nlitwin: That all sounds good to me! If we have `React.FC<Props>`, then we can omit the second `: Props`

frontend/pages/blog/introducing-journaly.tsx
nlitwin: wooo!!!!

frontend/pages/blog/introducing-journaly.tsx
nlitwin: good call

frontend/components/Layouts/BlogPageLayout.tsx
robin-macpherson: Sweet! 😄

frontend/components/Layouts/BlogPageLayout.tsx
robin-macpherson: Yep I meant to edit that out this morning haha

Issue #31, opened 2020-03-31
Port remaining My Feed code over


robin-macpherson:
## Description

Now that the Dashboard has been ported over and the `feed` query is working on the client, it's time to port over the remaining Feed code!

## Before Filing

- [x] I have checked for duplicate issues
- [x] I have filled out this template properly

PR #32, opened 2020-03-31
Port Remaining Feed Code Over

robin-macpherson:
## Description

**Issue:** #31 

Now that the Dashboard has been ported over and the feed query is working on the client, it's time to port over the remaining Feed code!

## Sub-Tasks

- [x] Port over the rest of our CSS Reset/Global Styles
- [x] Re-create and TypeScriptify `PostCard`
- [x] Finish re-creating `MyFeed` and polish everything up
- [x] Update route to `my-feed` instead of old `your-feed`


frontend/components/Dashboard/PostCard/PostCard.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react'
+import Link from 'next/link'
+
+interface IPostCardProps {
```
nlitwin, line 4: In other typescript/React apps, I've just done `interface Props {` since it's local to the file and isn't used elsewhere. What do ya think?

frontend/styles/globalStyles.css
```diff
@@ -42,15 +61,57 @@ a {
...
+
+input:focus {
+  outline: 0;
+}
+
+input,
+textarea {
+  -webkit-box-sizing: content-box;
+  -moz-box-sizing: content-box;
+  box-sizing: content-box;
```
nlitwin, line 104: How come `content-box` for all input types except the ones below?

frontend/styles/globalStyles.css
```diff
@@ -42,15 +61,57 @@ a {
...
+
+button,
+input[type='reset'],
+input[type='button'],
+input[type='submit'],
+input[type='checkbox'],
+input[type='radio'],
+select {
+  -webkit-box-sizing: border-box;
+  -moz-box-sizing: border-box;
```
nlitwin, line 115: We don't need vendor prefixes for `box-sizing`, they're supported even in IE8

frontend/styles/globalStyles.css
```diff
@@ -1,17 +1,31 @@
 html {
+  box-sizing: border-box;
```
nlitwin, line 2: This is in `reset.css` already

frontend/styles/globalStyles.css
```diff
@@ -1,17 +1,31 @@
...
   font-size: 10px;
 }
 
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
```
nlitwin, line 9: Same here

frontend/styles/globalStyles.css
```diff
@@ -1,17 +1,31 @@
...
+}
+
 body {
-  font-family: "Source Sans Pro", sans-serif;
+  padding: 0;
```
nlitwin, line 13: Same here for padding and margin

frontend/styles/globalStyles.css
robin-macpherson: ✅

frontend/styles/globalStyles.css
robin-macpherson: ✅

frontend/styles/globalStyles.css
robin-macpherson: ✅

frontend/styles/globalStyles.css
robin-macpherson: I just brought over what we had before. Changed them all to be unified ✅

frontend/styles/globalStyles.css
robin-macpherson: ✅

frontend/components/Dashboard/PostCard/PostCard.tsx
robin-macpherson: It's quite a common convention to prefix interface names with `I` so I was just establishing that across the app, but I suppose you're right that in the case of Props it might not matter as much since it's local and not used elsewhere. We can go with that for `Props` 👍🏼✅

Issue #36, opened 2020-04-03
Implement internationalization with i18n


robin-macpherson:
#### Perceived Problem

- By definition, we will have users from many countries and backgrounds who will speak many languages. Many of these people will either not speak English as a native language, or not at all.
- A lot of users will speak English well enough to navigate the web, in many cases out of necessity, but not as comfortably in their native or another language.

#### Ideas / Proposed Solution(s)

- We should build out infrastructure for internationalization so that as and when we are able to, we can provide a more comfortable experience for as many of our users as possible

PR #47, opened 2020-04-06
Implement i18n 🌍 in the app

nlitwin:
## Description

**Issue:** #36 

Support internationalization in the app! Follow-up PRs will include more translation namespaces (one for each page, for instance) and more usages of the English translations.

### Using the `next/error` page with custom translation
image

## Sub-Tasks

- [x] Update the docs
- [x] Convert `_app.js` and `_document.js` to `.tsx`
- [x] Add required namespaces to each page


frontend/pages/dashboard/my-feed.tsx
```diff
@@ -1,26 +1,26 @@
...
-
+import { useTranslation } from '../../config/i18n'
 import DashboardLayout from '../../components/Layouts/DashboardLayout'
 import MyFeed from '../../components/Dashboard/MyFeed'
-
 import { useFeedQuery } from '../../generated/graphql'
 
 interface InitialProps {}
 
 interface Props extends InitialProps {}
```
robin-macpherson, line 10: Shouldn't we add something to our Initial props interfaces on these pages now?

frontend/README.md
```diff
@@ -1 +1,43 @@
...
+```jsx
+import React from 'react'
+import { useTranslation } from '../config/i18n'
+
+export function MyComponent() {
+  const { t } = useTranslation()
+
+  return <p>{t('homePage.welcomeMessageTitle')}</p>
+}
+```
```
robin-macpherson, line 43: Thanks for adding the docs! 😄

frontend/config/i18n.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports = new NextI18Next({
+  defaultLanguage: 'en',
+  // Other languages will be added here once there are translations for them, e.g. ['es', 'de']
+  otherLanguages: ['de'],
+  // The "dev" fallback language will show the translation key, indicating there is a missing translation
+  fallbackLng: process.env.NODE_ENV === 'production' ? 'en' : 'dev',
+  detection: {
+    order: ['navigator', 'cookie'],
+  },
+})
```
robin-macpherson, line 12: Is there any particular reason to not make this a `.ts` file?

frontend/README.md
nlitwin: my pleasure!

frontend/config/i18n.js
nlitwin: I tried making this and the server file both TypeScript files, but ran into some dead ends. Making the server file into ts required quite a few new libraries and it just didn't seem worth it for just one file. Then because the server file imports this file and is in the node environment/uses CommonJS, it requires the exports of this file to be different, and it had trouble when it was `.ts`

frontend/pages/dashboard/my-feed.tsx
nlitwin: Hmm, that's a good question... Right now `namespacesRequired` isn't ever used within the component, but I think there are use cases when you do need to reference one. I'll play around with it and test it out!

robin-macpherson: Awesome sounds good! 😊

frontend/config/i18n.js
robin-macpherson: Ok we can leave it like this for the time being 😊

frontend/pages/dashboard/my-feed.tsx
nlitwin: I added `namespacesRequried: string[]` to `InitialProps`, but TypeScript doesn't like that Props inherits it without adding anything, so I just made it `<Props>`

robin-macpherson: Cool, this feels better to me, thanks for looking into it 🙂

Issue #53, opened 2020-04-11
Inline Feedback Experience, PART 1 - Text Selection and Highlight on "Comment"


robin-macpherson:
**Part 1 of #52 - Handle Text Selection and Highlight on clicking a "Comment" button.

PR #54, opened 2020-04-12
FEATURE: Inline Feedback Experience - Part 1

robin-macpherson:
## Description

**Issue:** #53 

### Background
- Writing comments in a standard way at the bottom of a post and one-by-one independently of each other is not an optimal or pleasant experience for neither the commenter not the reader.
- We need to make it a delightful experience to leave someone feedback and also to potentially interact with them back-and-forth; as well as for the person receiving the comments when we consider they will want to read the comments and reference the text where it was left.

### Feature
Build an Inline Feedback Experience where users are able to leave feedback on someone's writing in the exact place where they have a comment, and the Journaler will be able to comfortably read that comment right next to where it was left.

#### Part 1 (this PR)
- First build out the initial feature of being able to highlight text and click on a "comment" button which will highlight the selected area. This will allow us insight into the mechanism we will have to expand upon to store these inline comments in the DB and so on.

## Video

inline-feedback-experience-p1

## Sub-Tasks

- [x] Figure out the basic mechanism for selecting text on a page
- [x] Build a very rough POC for making Text Selection work and highlighting that text
- [x] Simple POC for working with SVGs
- [x] Handle initial edge cases
- [x] Handle commenting restrictions (see description)
- [x] Handle remaining edge cases
- [ ] Clean up code
- [ ] TypeScriptify
- [ ] Write documentation


frontend/components/Dashboard/Post/Post.tsx
```diff
@@ -1,12 +1,137 @@
 import React from 'react'
 import Head from 'next/head'
+
 import { Post as PostType } from '../../../generated/graphql'
+import { brandBlue, highlightColor, darkGrey } from '../../../utils'
+import IconLeaveComment from '../../Icons/IconLeaveAComment'
```
nlitwin, line 6: I hope this was an intentional pun 😄

Issue #55, opened 2020-04-18
Design & Implement a system for using SVGs at Journaly


robin-macpherson:
#### What

- We should create a nice system and convention for how we work with SVGs at Journaly.

#### Why

- At each company I've worked at so far, I've found that there has been no defined system or set of conventions for how they work with SVGs, which always results in people doing it any old way, makes it hard for new devs to understand how they should work, and ends with sub-par results in terms of code quality and performance.

#### How

- As a part of PR #54 I began building out a sort of POC for how we could create a delightful system where we use the [SVGR CLI](https://react-svgr.com/docs/cli/) to convert SVGs to React components. I will describe this in more detail in the approach doc in PR<TBD>.

PR #56, opened 2020-04-18
Approach Doc - Design & Implement a system for using SVGs at Journaly

robin-macpherson:
## Description

**Issue:** #55 

Approach doc to Design & Implement a system for using SVGs at Journaly.

## Sub-Tasks

- [x] Write first draft
- [x] Get Nick's thoughts
- [x] Final draft, merge, off to the workshop!

### DIFF
--- a/docs/approach-docs/system-svg.md
+++ b/docs/approach-docs/system-svg.md
@@ -1,2 +1,17 @@
 # A System For Working With SVGs - Approach Doc
 
+I would like to propose that we establish a delightfully simple and performant system for working with SVGs at Journaly.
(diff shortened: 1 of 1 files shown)


nlitwin:
@robin-macpherson this looks great to me! Thanks for the helpful video links in the markdown file 😄  The system I've used before only does like 1/3 of what `svgr` does, so I think we should use this approach!

As for integrating new SVGs into the repo, I think we could:

1. Run `svgr` on the new SVG
2. Commit the component that is outputted
3. Discard the original (?) - I don't think we'll need the originals in `public/images/icons/*`

It looks like there are some additional options for `svgr` to output TypeScript as well 👍

robin-macpherson:
@nlitwin glad you like it! That approach sounds good to me 😄 Let's merge this in then ✌🏼

Issue #39, opened 2020-04-05
Use typescript-eslint since tslint is deprecated


nlitwin:
Swap out `tslint` for `typescript-eslint`, since `tslint` is deprecated

More React/TypeScript resources for setting things up:
- https://www.sitepoint.com/react-with-typescript-best-practices/
- https://github.com/typescript-cheatsheets/react-typescript-cheatsheet

PR #60, opened 2020-04-27
Replace deprecated tslint with typescript-eslint

nlitwin:
## Description

Resolves: #39, replacing `tslint` with `typescript-eslint`. I resolved most TypeScript errors except for some complicated files that had a lot going on and would require a deep dive. For now, they are ignored with `// @ts-nocheck`

## Sub-Tasks

- [x] Resolve TypeScript errors


frontend/components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -1,10 +1,11 @@
+// @ts-nocheck
 import React from 'react'
 
 import PostCard from '../PostCard'
 
 // import { Post } from '../../../generated/graphql' // TODO (robin-macpherson): See Post[] type todo below
 
-interface Props {
+type Props = {
```
nlitwin, line 8: @robin-macpherson I wanted to get your feedback on this one before I made a lot of updates. After some research, I'm thinking that types for props as opposed to interfaces is better for React components. A very popular [React TypeScript cheatsheet](https://github.com/typescript-cheatsheets/react-typescript-cheatsheet#types-or-interfaces) with 12.5k github stars recommends types for props and state because its much better for unions between types, and in my other React projects where we're using interfaces, I've found them to be a bit harder to work with. Let me know whatcha think 😊

frontend/components/Icons/IconLeaveAComment.tsx
```diff
@@ -1,13 +1,21 @@
...
+import React from 'react'
 
-function SvgComponent({ size, primaryColor, secondaryColor }) {
+type Props = {
+  size: number
+  primaryColor: string
+  secondaryColor: string
+}
+
+function SvgComponent(props: React.SVGProps<SVGSVGElement> & Props) {
```
nlitwin, line 9: Here's an example combining the type `Props` with `SVGProps`

frontend/components/Dashboard/MyFeed/MyFeed.tsx
nlitwin: Here's another good [article](https://medium.com/@martin_hotell/interface-vs-type-alias-in-typescript-2-7-2a8f1777af4c) on types vs interfaces

robin-macpherson: Thanks for the cheatsheet and article, @nlitwin ! This makes sense and I'm fine with moving forward with this approach 😊💪🏼

PR #62, opened 2020-05-03
Create script for converting SVGs to React components

nlitwin:
## Description

**Issue:** #55 

### Command prompts

svgr_prompts

### Error handling

svgr_prompts_errors

## Sub-Tasks

- [x] Write script for converting svg files into React components
- [x] Add docs
- [x] Handle a11y


frontend/scripts/svgr.js
```diff
@@ -0,0 +1,99 @@
...
+  },
+  {
+    type: 'text',
+    name: 'componentName',
+    message: 'What do you want the SVG component to be called?',
+    validate: (input) => {
+      if (input.includes('.')) {
+        return "Pleaes omit the file extension when naming your component. 'Circle' will generate Circle.tsx"
+      } else if (input.toLowerCase().endsWith('icon')) {
+        return "Do not include 'Icon' in the component name"
```
nlitwin, line 80: Once I started using the component, like in `NavLinks.tsx`, I realized it might actually be nice to always include `Icon` in the file name. Then you could call it `Feed` and it would generate `FeedIcon.tsx`. What do you think?

robin-macpherson: Yep I definitely think that `Icon` should be in the name. I think the idea of adding it automatically with the script is nice! Let's do it :)

frontend/components/Dashboard/Nav/NavLinks.tsx
```diff
@@ -35,13 +36,13 @@ const NavLinks: React.FC<Props> = ({ onClick }) => {
...
           <a className="nav-link" onClick={onClick}>
-            <img src="../images/icons/your-feed-icon.svg" alt="" />
+            <Feed aria-hidden="true" />
             <span className="nav-link-text">{t('nav.myFeed')}</span>
           </a>
         </NavLink>
         <NavLink href="/dashboard/my-posts">
           <a className="nav-link" onClick={onClick}>
-            <img src="../images/icons/your-feed-icon.svg" alt="" />
+            <Feed aria-hidden="true" />
```
robin-macpherson, line 45: Yeah these definitely need `Icon` in the component name

frontend/scripts/svgr.js
```diff
@@ -0,0 +1,99 @@
+const fs = require('fs')
+const path = require('path')
+const svgr = require('@svgr/core').default
+const prompts = require('prompts')
+// https://gist.github.com/pladaria/69321af86ce165c2c1fc1c718b098dd0
+const svgoConfig = {
+  plugins: [{ removeViewBox: false }],
+}
```
robin-macpherson, line 8: Did you also want to add the `{ "cleanupIDs": false }` that was in the original `svgo.json` config?

nlitwin: When I was testing it, enabling `cleanupIDs` seemed to help clean up the SVGs but not affect how the SVG was rendered. Was the original value in `svgo.json` preventing a bug or something that I maybe missed?

nlitwin: Merging for now, but will follow up with a quick PR if `cleanupIDs` is needed

robin-macpherson: Yeah the whole point of the settings I added to the `svgo.json` was to help clean up and optimize the SVG output, in part so that they are easier to work with, especially if in some cases there are animations or all kinda of things we might want to manipulate in the SVG. In general I think it's a good idea to check with each other in a case like this before removing something one of us put there for a reason, but it's not a huge deal of course.

nlitwin: We might be saying the same thing, but I removed `cleanupIDs: false`, so that that the default of true would be enabled, meaning that the SVG would be cleaned up even more. I checked the the previous SVGs to make sure I didn't break anything.

Issue #59, opened 2020-04-26
Proposal for modals


nlitwin:
For modals in our React app, we can make use of [React Portals](https://reactjs.org/docs/portals.html). This allows creating a separate "modal-root" element in the DOM, which is a sibling of the application root.

```html
<body>
  <div id="root" />
  <div id="modal-root" />
</body>
```

The parent `<Modal />` component can simply render `children` inside of it and all of it will be rendered in `div#modal-root`.

```jsx
<Modal>
  <div>I am the modal content</div>
</Modal>
```

The `<Modal />` component can be very minimal, with simple wrapping divs that handle animations, the blurred out background, and a close button. It will also handle rendering the modal in the center of the screen. The size of the modal and the content itself can be determined by the component passed in as `children` to `<Modal />`.

The main benefit to using React Portals is that the modal is not rendered as children to the content that triggers opening it. For example, your modal shouldn't need to be rendered in the DOM here:

```jsx
<Page>
  <button onClick={() => setIsModalOpen(true)}>Open modal</button>
  {isModalOpen && <Modal><div>Modal content</div></Modal>}
</Page>
```

robin-macpherson:
Sounds great to me! Proposal ✅👌🏼 Let's put this in an approach doc as per the pattern I've set up for now so that we have them in version control.

PR #63, opened 2020-05-04
Create a modal component

nlitwin:
## Description

Resolves: #59 

Fully accessible modal, full screen on mobile, centered on desktop. See gifs below for UX

## Sub-Tasks

- [x] Animates title and keeps header/footer fixed when there is a lot of scrollable content
- [x] Implements [a11y UX pattern](https://www.w3.org/TR/wai-aria-practices/#dialog_modal) for modals
  - [x] tabbing stays within the modal
  - [x] esc key closes the modal
  - [x] the rest of the app is not interactable when modal is up
  - [x] after you close the modal, focus returns to the button that originally opened the modal
- [x] Uses [React Portals](https://reactjs.org/docs/portals.html) so `<Modal />` components don't have to be rendered as children to content that's only responsible for opening it
- [x] Create `<Button />` component that can be unstyled (useful for wrapping icons but keeping good semantics). We can enhance it over type with different sizes, default styles, primary, secondary, error states, disabled etc.

## Example usage

```jsx
const Component = () => {
  const [openModal, setOpenModal] = useState(false)
  const triggerElementId = 'open-modal-button'
  const handleClick = () => setOpenModal(true)

  return (
    <Button id={triggerElementId} onClick={handleClick}>
      Open Modal
    </Button>

    {openModal && (
      <Modal
        onClose={() => setOpenModal(false)}
        title="This is the modal title"
        body={<div>Content of the modal</div>}
        footer={<div><button>Submit modal form</button></div>}
        triggerElementId={triggerElementId}
        maxHeight={'700px'}
      />
    )}
```
 

### Desktop

modal_desktop

### Mobile

modal_mobile

### DIFF
--- a/frontend/.prettierrc.js
+++ b/frontend/.prettierrc.js
@@ -2,5 +2,5 @@ module.exports = {
   semi: false,
   singleQuote: true,
   trailingComma: 'all',
-  printWidth: 80,
+  printWidth: 100,
 }
--- a/frontend/components/Hooks/useFocusTrap.ts
+++ b/frontend/components/Hooks/useFocusTrap.ts
@@ -1,0 +1,85 @@
+import { useEffect } from 'react'
+
+type Options = {
+  // The element that we want to keep focus within
+  rootElement: HTMLElement
+  // Callback for hitting the escape key
+  onClose: () => void
+  // Once the focus trap is done with, focus should return to the element that initiated it
+  returnFocusElementId?: string
+  // Class to add to the body while focus trapped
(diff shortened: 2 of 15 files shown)


nlitwin:
@robin-macpherson I wasn't sure about some of the styling (X close button, border colors, button   colors + hovers, etc.) so there is some work to be done there. Since we currently don't have any usage of the modal, I think this is ok, and once we have our theme/design languages more fleshed out, we can update it.

PR #67, opened 2020-05-16
Implement theme with typography, colors, and breakpoints

nlitwin:
## Description

POC for proposal #68 

🚨 Note: current typography font values are placeholders. They are not currently updated with our values

## Sub-Tasks

- [x] Simplified variable imports, just `theme` now
- [x] Use TypeScript for everything so autocomplete works
- [x] Allow breakpoints to be used as numbers as well as strings with `px` (numbers for allowing math operations, like + or - 1 scenarios, and `'px'` for regular media queries)

### DIFF
--- a/frontend/theme/index.ts
+++ b/frontend/theme/index.ts
@@ -25,7 +25,8 @@ type Colors =
   | 'red'
 
 type Typography =
-  | 'fontFamily'
+  | 'fontFamilySansSerif'
+  | 'fontFamilySerif'
   | 'paragraphSM'
   | 'paragraphMD'
   | 'paragraphLG'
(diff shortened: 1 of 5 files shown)


frontend/theme/index.ts
```diff
@@ -0,0 +1,103 @@
+import { fromEntries } from '../utils'
+
+export const breakpoints = {
+  XS: 600,
+  SM: 786,
+  MD: 1024,
+  LG: 1280,
+} as const
```
nlitwin, line 8: Our current breakpoints are 375, 960, 1040, and 1364, but I think we can tweak these a little bit to handle more use cases. For example, we may never need 375, especially since that's the smallest value we support and if we write "mobile first" CSS, we won't need it as a breakpoint. For example:

```css
.my-box {
  /* apply to every breakpoint, starting from 0px */
  color: tomato;
  @media (min-width: 800px) {
    /* override and apply to 800px and up */
    color: papayawhip;
  }
}
```

I also think that since our breakpoints of 960 and 1040 are really close together (only 80px apart), there will rarely be situations where we want to use both. How do you feel about the breakpoints I listed above?

600 allows us to target from 0 -> 600 (all phones) and then start switching to "tablet" layouts, if necessary
768 allows us to target tablets specifically (iPad having a width of 768)
1024 is a good middle ground between our current 960 and 1040 that is still divisible by 16
1280 is another pixel value that's good for grids and can be a starting point for "desktop and beyond"

frontend/theme/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  typography: {
+    readonly [key in Typography]: string
+  }
+}
+
+const theme: Theme = {
+  colors: {
+    black: '#393939',
+    white: '#ffffff',
+    gray100: '',
```
nlitwin, line 39: Since grays are frequently used for borders, backgrounds, text, etc., I think we'd greatly benefit from a range of values. I wasn't exactly sure what to put for these values, cause I wanted to include our current grays and then create the other ones based on those, but I might need your help for that.

What do you think?

frontend/theme/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    gray200: '',
+    gray300: '',
+    gray400: '',
+    gray500: '',
+    gray600: '',
+    gray700: '',
+    gray800: '',
+    blue: '#32567E',
+    blueLight: '#4391C9',
```
nlitwin, line 48: I really like this naming scheme because it allows for quite a few variations `blueDarkest, blueDarker, blueDark, blue, blueLight, blueLighter, and BlueLightest`. Lemme know what you think!

frontend/theme/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    red: '#c42f14',
+  },
+  breakpoints: fromEntries(
+    Object.entries(breakpoints).map(([key, value]) => [key, `${value}px`]),
+  ) as Breakpoint,
+  typography: {
+    fontFamily: '"Source Sans Pro", sans-serif',
```
nlitwin, line 55: Just a reminder that these are placeholder values, but I can create a screenshot of all these values shown on the screen in case we might want to use them?

frontend/utils/index.ts
```diff
@@ -0,0 +1,12 @@
+export * from './breakpoints'
+export * from './colors'
+export * from './css'
+export * from './date'
+
+// Polyfill Object.fromEntries
```
nlitwin, line 6: I was having trouble with next.js / TypeScript allowing me to use es2019 `Object.fromEntries` so I just polyfilled it

PR #72, opened 2020-05-17
Implement LoginForm with react-hook-form

### LINKED ISSUE #70: Proposal for using react-hook-form
#### Perceived Problem

When writing out form logic, I ran into the scenario where I needed more types of validations and hooks, as well as support for different input types like `type="file"`. We've made some great progress with `useFormValidation.ts` file, but we will need quite a lot more to handle all of our use cases. The time investment is also quite high if we choose to continue rolling our own solution.

After researching various form libraries, I think we would greatly benefit from [react-hook-form](https://github.com/react-hook-form/react-hook-form), given that it has a small footprint, built to be used with hooks, and integrates well with TypeScript. It also has amazing features like a [DevTools extension](https://react-hook-form.com/dev-tools) and even [a form builder](https://react-hook-form.com/form-builder) (!) where you tell it which types of fields and validations you want and it spits out the code to copy!

Since the docs are great and we wouldn't have to build much out on our own, we would probably end up saving at least around 30-40 hours of time by using the library.

A POC PR will accompany this proposal, which should have enough examples to show how it feels to use the library, and we can evaluate whether we want to use it.

nlitwin:
## Description

**Issue:** #70 

This is the POC for using `react-hook-form`. It's actually one of the most pleasant experiences integrating a form library that I've ever had 😌 Every single thing I wanted to do, I found in the docs, complete with multiple examples and codepens!

## Sub-Tasks

- [x] Display both form errors + apollo errors
- [x] Handle validation: required, pattern, and minLength
- [x] Handle form errors for each validation

### DIFF
--- a/frontend/elements/Button/Button.tsx
+++ b/frontend/elements/Button/Button.tsx
@@ -13,13 +13,21 @@ type Props = {
   id?: string
   variant?: Variant
   plain?: boolean
+  type?: 'button' | 'submit' | 'reset'
 }
 
 const Button: React.FC<Props> = (props) => {
-  const { children, onClick, plain = false, variant = Variant.Primary, ...otherProps } = props
+  const {
+    children,
+    onClick,
+    plain = false,
+    variant = Variant.Primary,
+    type = 'button',
+    ...otherProps
+  } = props
 
   return (
-    <button onClick={onClick} {...otherProps}>
+    <button onClick={onClick} type={type} {...otherProps}>
       {children}
 
       <style jsx>{`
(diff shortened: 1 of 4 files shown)


frontend/elements/Button/Button.tsx
```diff
@@ -13,13 +13,21 @@ type Props = {
   id?: string
   variant?: Variant
   plain?: boolean
+  type?: 'button' | 'submit' | 'reset'
```
nlitwin, line 16: Temporary changes on the button component until #71 is merged

PR #73, opened 2020-05-29
Replace generic `<Form />` component with form tailored for settings

nlitwin:
## Description

**Issue:** #74 

## Subtasks

- [x] Delete `<Form />` in favor of a form component tailored for the settings pages

## Screenshots

⬇️  Shows how the forms / fieldsets will work for each additional section. Here `Your Details` is one form and `Languages` is another

image


frontend/components/Modal/ModalHeader.tsx
```diff
@@ -13,7 +13,7 @@ interface Props {
 const ModalHeader: React.FC<Props> = ({ onClose, title, showTitle }) => {
   return (
     <div className="modal-header">
-      <Button plain onClick={onClose} className="modal-close-button">
+      <Button icon onClick={onClose} className="modal-close-button">
```
nlitwin, line 16: Somehow missed this prop rename in a previous PR

frontend/pages/dashboard/settings/profile.tsx
```diff
@@ -2,35 +2,64 @@ import React from 'react'
...
-            <img className="profile-image" src="/images/robin-small.png" />
-
-            <Button className="details-submit-button" variant={Variant.Secondary}>
-              {t('profileForm.submitButton')}
-            </Button>
-          </div>
-        </fieldset>
-      </Form>
+      <div className="forms-container">
+        <SettingsForm onSubmit={handleDetailsSubmit}>
```
nlitwin, line 19: What do you think of this structure? Since there will be about 5 sections/forms on the profile settings page, each with their own `Update`/submit button, I split them into multiple forms. It's probably possible to make one form technically, and then have multiple submit buttons within the single form, but this will depend on the Apollo/mutation setup. I'm open to structuring it both ways (1 form, multiple submit buttons / multiple forms, one submit button for each). I'm kind of thinking it makes sense to just start like this and if it makes sense to change it later depending on the Apollo stuff, I can make that change pretty quickly

robin-macpherson: I'm down with starting like this and seeing if we need the change later 😄

Issue #88, opened 2020-06-14
Inline Feedback Experience, PART 2 - Load Post With Persisted Comments


robin-macpherson:
#### Description

**Part 2 of #52  - Handle ability to view a post and load persisted comments and reconstruct the post.

#### Considerations

- This system should work well with both being able to load comments with the exact selection and the position they were made in, while also allowing for new comments to be added without breaking the positioning of existing comments.

PR #80, opened 2020-05-31
Update types for Posts and highlighting text

nlitwin:
## Description

**Issue:** #77 

## Subtasks

- [x] I have added this PR to the Journaly Kanban project ✅
- [x] Update types
- [x] Use the theme

PR #89, opened 2020-06-14
FEATURE: Inline Feedback Experience PART 2 - Handle Loading Persisted Comments

robin-macpherson:
## Description

**Issue:** #88 

Handle loading a post with existing comments and being able to reconstruct the post with all highlights and comment threads in the exact position they were left in with the same text selection.

- We built this in a way that allows us to perfectly position persisted/exisitng comments and their text selections upon visiting a post, while not breaking that positioning when adding new comments.
- We achieved this by anchoring comment thread locations to the beginning of the document as a single source of truth.
- Some refactoring of re-used logic into functions and better function/variable names
- Added documentation/comments

## Subtasks

- [x] Write mechanism for correctly storing a comment's position and selection and loading them in the correct place
- [x] Handle conflicts with old and new comment positioning
- [x] Test lots of scenarios
- [x] Clean up and refactor
- [x] Add documentation with comments

## Screenshots
persisted-comments

### DIFF
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -1,6 +1,6 @@
 datasource db {
   provider = "postgresql"
-  url      = "postgresql://jlyuser:jlypass@localhost:5433/jlydb"
+  url      = env("DATABASE_URL")
 }
 
 generator prisma_client {
--- a/frontend/components/Dashboard/Post/Post.tsx
+++ b/frontend/components/Dashboard/Post/Post.tsx
@@ -11,6 +11,7 @@ interface IPostProps {
   post: PostType | any
 }
 
+// Elements whose boundaries a comment can cross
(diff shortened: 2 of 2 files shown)


robin-macpherson:
@nlitwin woohooooooo this one is so excoitingggggggggg 🤩🪂☀️

Lanny:
#93 which is based off this branch refactors things significantly so merging #80 will lead to some tricky merge conflicts for that. Just a heads up though, totally agree that we probably should have been typing things from the start and we should add them sooner than later

nlitwin:
Ahh I see, ya merging in #80 would cause #93 some issues. If one of you ends up adding types to `Dashboard/Post/Post.tsx`, feel free to copy over any types from #80. I'm happy to just close/delete that PR

robin-macpherson:
@nlitwin @Lanny WOOHOOOOOO!!!!!!

PR #92, opened 2020-06-14
Wire Up User Login & `currentUser` query

robin-macpherson:
## Description

**Issue:** #99 

- This PR implements login functionality as well as wiring up the `currentUser` query.
- Also adds conditional rendering of the `NavLinks` depending on whether the user is logged in or not.
- Added `profileImage` to `User` model
- Includes migration

login-redirect-refetch

## Subtasks

- [x] Write backend Query & Mutation
- [x] Update `LogInForm`
- [x] Debug cookie issues
- [x] Update datamodel with `profileImage` for `User`s
- [x] Create/Apply DB migration
- [x] Update `.graphql` files on frontend and generate Apollo hooks & types


backend/src/app.ts
```diff
@@ -22,6 +22,7 @@ server.express.use(cookieParser())
 // decode JWT to be used on each request
 server.express.use((request: any, response, next) => {
   const { token } = request.cookies
+  console.log('token', token)
```
robin-macpherson, line 25: @Lanny ⬆️🙋🏻‍♂️

Lanny:
@robin-macpherson It seems to work for me, I added `useCurrentUserQuery()` to `Dashboard.tsx`, on server side I can see the token printed out and I'm getting back user info for the user I signed up as, across multiple singups.

How are you testing?

I did notice the cookie isn't being sent in the playground app thing that runs on port 4000, likely because when you sign up Apollo orchestrates setting the cookie for the current origin, meaning localhost:3000 in local, this isn't carried over to localhost:4000 because it's a different origin. But in the context of actually looking a Journaly pages it seems to work as expected

backend/src/graphql.ts
```diff
@@ -191,7 +190,42 @@ const Mutation = mutationType({
         })
         return user
       },
-    })
+    }),
+      t.field('loginUser', {
```
Lanny, line 194: I don't know if we have tabs mixed in on this file, but the indentation always looks different (and wrong) on github.

frontend/components/Dashboard/Nav/NavLinks.tsx
```diff
@@ -28,8 +31,8 @@ const NavLinks: React.FC<Props> = ({ onClick }) => {
       <div className="nav-top">
         <Link href="/dashboard/profile">
           <a onClick={onClick}>
-            <img className="profile-img" src="/images/robin-small.png" />
-            <p className="current-user-name">Robin MacPherson</p>
+            <img className="profile-img" src={currentUser.profileImage} />
```
Lanny, line 34: Since this is nullable, this should probably be `{currentUser.profileImage && (<img....`

robin-macpherson:
@Lanny finally got the redirect and refetching to work after login! See GIF 😀

Lanny:
Awesome! It's coming to life! 🐚

frontend/components/Dashboard/Nav/NavLinks.tsx
robin-macpherson: Ah, good point. I think I may have another idea I like a wee bit better, though! Just a moment

backend/src/graphql.ts
robin-macpherson: Yeah I've noticed that, too. Almost mentioned it earlier!

frontend/components/Dashboard/Nav/NavLinks.tsx
robin-macpherson: Voilà !! Whatcha think?

image

Lanny: Looks awesome! If you want to make it reusable and take a size prop, we could start adding profile images next to comments too, although that can happen as part of a separate PR if you want.

robin-macpherson: Yeah that would be amazing! Let's do that in a separate PR, I was thinking of making another PR to fix up some of the UI styling that's gone a bit wonky (the color of the dashboard nav and header elements, for example) so perhaps could just include some other improvements in there, too! 

Merging in for nowwww. Grab your things hedgehog!  🏎🧵...🦔🧳💨

PR #100, opened 2020-06-21
Create Launch Database Seed Script For Languages & Topics

robin-macpherson:
## Description & Problem Statement

I've been thinking a lot about the issue of how we go about adding topics and languages for quite some time now. @nlitwin and I discussed this a number of times and I feel like it's a very tricky issue. For example, do we create them ahead of time, or do we allow users to create them? 

It seems like it would be nice to let users add a new language or topic, but then how do we prevent people from creating "Chinese" and "Mandarin" and "Mandarin Chinese" and "chinese" and maybe even "cjinese"? Not to mention with the topics!

### Languages

With languages, I feel it's actually a bit simpler, since there are technically a finite number of them. In theory, we can simply add them all and allow people to type a search, updating it with the matching results, and maybe even create some organization by region, etc. in the menu.

### Topics

This one is a little more tricky, given how many topics and variations we could have. For now, I'm still thinking one option is to just think of an enormous amount of topics to where people could surely find one that sort of matches?

## Proposal

Regardless of where we land on the above questions, I'm thinking that it would be nice either way to be able to run a script after we launch that creates a huge selection of languages and topics in the database, ready to go.

The idea here would be that we can run this once right after we launch, but it could also help serve us during development. It's nice to have it separate because we don't need to create seed users and stuff on the launch, and this helps us preserve our nice Game Of Thrones themed seed script but we can still use this if we want a ton of languages and topics in local dev.

This is a first draft just to show the concept, and I can continue adding to it as a living document if everyone likes the approach!

## Subtasks

- [x] Create first draft
- [ ] If approved, add remaining languages
- [ ] ^ add tons of topics

### DIFF
--- a/backend/launch.ts
+++ b/backend/launch.ts
@@ -1,0 +1,316 @@
+import { PrismaClient } from '@prisma/client'
+
+const db = new PrismaClient()
+
+main().catch((e) => {
+  throw e
+})
+
+async function main() {
+  // Create base set of languages
+
+  await db.language.create({
+    data: {
+      name: 'Mandarin Chinese',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Spanish',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'English',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Hindi',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Bengali',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Portuguese',
+      dialect: 'European',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Portuguese',
+      dialect: 'Brazilian',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Russian',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Japanese',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Western Punjabi',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Marathi',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Telugu',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Turkish',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Korean',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'French',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'German',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Vietnamese',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Tamil',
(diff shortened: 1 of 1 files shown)


backend/launch.ts
```diff
@@ -0,0 +1,316 @@
+import { PrismaClient } from '@prisma/client'
+
+const db = new PrismaClient()
+
+main().catch((e) => {
+  throw e
+})
+
+async function main() {
+  // Create base set of languages
```
robin-macpherson, line 10: These are in order of population of speakers.

backend/launch.ts
```diff
@@ -0,0 +1,316 @@
...
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Gujarati',
+    },
+  })
+  await db.language.create({
+    data: {
+      name: 'Farsi',
```
robin-macpherson, line 132: This one brings up a good question of variations. Farsi is also widely known as Persian, although Farsi is the official name of the language today. It could be helpful to have a field for "variations"? 🧐

backend/launch.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  // Create topics
+  await db.topic.create({
+    data: {
+      name: 'Cooking',
+    },
+  })
+  await db.topic.create({
+    data: {
+      name: 'Rock Climbing',
```
robin-macpherson, line 189: This made me think of a similar issue with variations. @nlitwin and I have talked before about "topics" and then "subtopics". 

At the end of the day, I'd love for us to figure out what the simplest implementation and system here is...

One idea is similar to my thought above: maybe we have `name: "Climbing"` and `variation: "Rock"` (😂). 

Or just how I have done it here with "Rock Climbing" and "Bouldering" and so on as separate topics.

These could essentially be used as "tags" where you add all that apply.

backend/launch.ts
```diff
@@ -0,0 +1,316 @@
...
+    },
+  })
+  await db.topic.create({
+    data: {
+      name: 'Mathematics',
```
robin-macpherson, line 194: Wanted to put "Maths"! 🇬🇧💂🏼‍♀️

backend/launch.ts
Lanny: I think allowing user entry is going to be important here, I don't think we're going to ever come up with an exhaustive list of human interests. We can do some lexical normalization, e.g. standardizing "Rock climbing", "rockclimbing" and "Rock-Climbing" but "Bouldering" is going to require some semantic understanding of the interest. I think having an "topic alias" table might make sense, so we have one canonical "Rock Climbing" row in topics, but we consult the alias table to find out "Boldering" is another name for the same thing. Topics can be merged and aliased as people create new topics to try and get as much activity on each topic as possible

Lanny:
oh, topic name translation is a good point. Fixed set of topics is probably the only sane way to do it. Aliasing to help with searchability may still be desirable, but probably not urgent if it's a closed set.

robin-macpherson:
@nlitwin @Lanny thanks all for the input! I think this one is going to require some good discussion. Also some great opportunity for ML stuff here! 😀

PR #101, opened 2020-06-22
Expand Seed Data

robin-macpherson:
## Description

- Expand seed data to be a wee bit more robust
  - Add some extra posts and comments
  - Add images to the posts so that they will look nice and have the properties they need when rendered in the app via the feed, profile, etc.

...

## Subtasks

- [x] Add extra post with comment
- [x] Add an image to the seed script
- [x] Test the seed script - WORKS!


backend/prisma/schema.prisma
```diff
@@ -184,13 +184,12 @@ model Thread {
...
-  postId             String?
-  imageRole          ImageRole
-  Post               Post?     @relation(fields: [postId_ImageToPost], references: [id])
-  postId_ImageToPost Int?
+  id        Int       @default(autoincrement()) @id
+  smallSize String
+  largeSize String
+  imageRole ImageRole
+  Post      Post?     @relation(fields: [postId], references: [id])
+  postId    Int?
```
robin-macpherson, line 192: @Lanny it seemed like it should have been like this, right?

backend/prisma/schema.prisma
```diff
@@ -77,6 +77,7 @@ model Post {
   title      String
   body       String
   status     PostStatus   @default(DRAFT)
+  // @Lanny ⬇️
   images     Image[]
```
robin-macpherson, line 81: *Part 1
@Lanny so this column doesn't seem to show up in the DB when I view it and we're also unable to query it. Do we need to annotate the relation here? We shouldn't need to right because we name it on the `Post` model... I think I might be a wee bit confused here. See *Part 2 below

backend/seed.ts
```diff
@@ -160,6 +231,56 @@ async function main() {
...
+                    connect: { id: ned.id },
+                  },
+                },
+              ],
+            },
+          },
+        ],
+      },
+    },
+  })
```
robin-macpherson, line 266: @Lanny I have a wee bit of a strange situation when rendering this post in the app. It loads fine, and if I inspect the DOM I can see the comment thread appears to have been correctly applied around the selection, but we don't get the highlight for some reason 🧐. Any ideas?

backend/seed.ts
```diff
@@ -160,6 +231,56 @@ async function main() {
...
+      largeSize:
+        'https://res.cloudinary.com/journaly/image/upload/v1574187295/journaly/p5dxflsivmpwii9146rw.jpg',
+      imageRole: 'HEADLINE',
+      Post: {
+        connect: {
+          id: 1,
+        },
+      },
+    },
+  })
```
robin-macpherson, line 282: I had started making some updates to other code to try to accommodate actually showing these images in the app, but I ran into a few road blocks and it also falls perfectly into the UI updates PR I'm going to do next, so I decided to revert those changes for now and just keep this nice and specific to updating the seed data and will revisit the rest in that next PR

backend/seed.ts
```diff
@@ -160,6 +231,56 @@ async function main() {
     },
   })
 
+  // Add comment to Jon's post
+  await db.post.update({
+    where: {
+      id: 2,
```
Lanny, line 237: Using a hardcoded ID here probably isn't ideal, it could change based on when the seed data is loaded, or the order in which things are added in the seed data. The fastest way to add add threads is in the same prisma call that creates the post (same way a thread is created for the other one), but if you do want to make this thread separately you'll want to save the post ID above when it's created and reference that.

backend/seed.ts
Lanny: hmm, that's odd. Does it have the expected class name? `highlightedContent` doesn't seem to line up with the indices, how did you get the `startIndex` and `endIndex` values?

robin-macpherson: The indices should be correct. I actually got them by directly console.logging the `startIndex` and `endIndex` values while creating a text selection manually. It did appear to have the correct class name

backend/seed.ts
robin-macpherson: Yeah I couldn't do it in the same place as creating the post because that would mean that the other users weren't created yet and so it gave errors for referencing vars before they were declared

robin-macpherson: There we go. I changed the structure a little so that we start by creating the users, and then create the post afterwards along with comment threads and that way the users are all created and available to interact! 😄

PR #102, opened 2020-06-22
Add `excerpt` field to Post & make field name casing consistent

robin-macpherson:
## Description

@Lanny since you're going to be working on creation of posts next, while I was working on some things I realized that we should probably have an `excerpt` field on the post that we could construct when saving a `Post`, so that this can be displayed in the `PostCard` and not a string of HTML.

What do you think?

I also took this opportunity to quickly fix a little bit of remaining inconsistency with our field name casing.

Lastly, made a fix to Image field/type, see comment 🙂

## Subtasks

- [x] Add `excerpt` field
- [x] Fix inconsistent field name casing
- [x] Fix image relation stuff (?)


backend/prisma/schema.prisma
```diff
@@ -76,6 +76,7 @@ model Post {
   id         Int          @default(autoincrement()) @id
   title      String
   body       String
+  excerpt    String
```
robin-macpherson, line 79: Here's the new `excerpt` field ⬆️

backend/prisma/schema.prisma
```diff
@@ -184,13 +185,12 @@ model Thread {
...
-  postId             String?
-  imageRole          ImageRole
-  Post               Post?     @relation(fields: [postId_ImageToPost], references: [id])
-  postId_ImageToPost Int?
+  id        Int       @default(autoincrement()) @id
+  smallSize String
+  largeSize String
+  imageRole ImageRole
+  post      Post      @relation(fields: [postId], references: [id])
+  postId    Int
```
robin-macpherson, line 193: @Lanny it feels like it should have been like this, right? I think Prisma autogenerated that previous one perhaps due to us not switching the previous `postId` from `String` to `Int` perhaps?

Lanny: Yeah, posts have integer IDs so `postId` should definitely be an int. The string type was probably a holdover from uuids like you said

backend/prisma/schema.prisma
Lanny: Looks good, I'm guessing the excerpt be like the first N characters of the post (maybe non-heading characters)?

robin-macpherson: Yep exactly, that was my thought! Probably first N chars of the actual body content. In theory, the title of the post is on the header/image, so posts will likely start with body copy anyway, but will be good to have a check for that just in case since it's totally possible to do that with the rich text editor!

robin-macpherson:
Oh @nlitwin this will also help your work on the profile/my posts/drafts pages! You'll now be able to use the `post.excerpt` rather than just the body.

PR #105, opened 2020-06-24
Expand Commenting CRUD Operations : Backend

robin-macpherson:
## Description

**Issue:** #113 

This PR covers the backend portion of expanding the commenting CRUD operations and updates the necessary queries/mutations on the frontend for later. I have chosen to have this one light and break the remaining frontend work into a separate PR because we need to handle #112 first rather than adding `currentUser` queries to a lot of places.

## Subtasks

- [x] Implement `updateComment & `deleteComment` resolvers
- [x] Update graphql Query & Mutation files


backend/src/graphql.ts
```diff
@@ -301,5 +301,52 @@ schema.mutationType({
...
+        const { userId } = ctx.request
+        if (!userId) throw new Error('You must be logged in to do that.')
+
+        const comment = await ctx.db.comment.update({
+          data: {
+            body: args.body,
+          },
+          where: {
+            id: args.commentId,
+          },
```
Lanny, line 320:
```suggestion
          where: {
            id: args.commentId,
            authorId: userId
          },
```

backend/src/graphql.ts
```diff
@@ -301,5 +301,52 @@ schema.mutationType({
...
+      },
+    })
+    t.field('deleteComment', {
+      type: 'Comment',
+      args: {
+        commentId: intArg({ required: true }),
+      },
+      resolve: async (parent, args, ctx) => {
+        const { userId } = ctx.request
+        if (!userId) throw new Error('You must be logged in to do that.')
```
nlitwin, line 337: I wonder how we can translate these 🤔 I think we'll need to add the user's locale to the `context` and use it to retrieve the appropriate translation. So we may also have to move our locale files up from `/journaly/frontend/public/static/locales/**/*.json` into `/journaly`, then `/journaly/backend` could access them too. I created #106 to track and explore this

robin-macpherson: It might still be helpful to have our backend errors in English at the point that they are thrown in things like server logs, etc. so I wonder if we could create a simple mechanism where we have a list of these errors with translations on the frontend, and we send some kind of code along with the error and then when we catch it on the frontend, map it to the correct error that has translations set up? Just thinking outloud

robin-macpherson: It would be a bit of a conundrum for Ryan to be digging into our server logs in the middle of the night trying to fix a production fire and they're all in Punjabi, Indonesian, and Hakka Chinese 🧐 😂

robin-macpherson: I feel like should be able to have a pretty manageable list of client-side errors that we need to display, so that's why I'm thinking something like that could perhaps work well. But open to other ideas. Great to have the discussion here but I think we could handle this in its own PR 😊( unless we come up with a solution pretty easily in the next day or so)

nlitwin: For sure! I was _just_ about to comment that we could also just send an error code from the backend and have that be the key to the translation. We'd just want to make sure to maintain one source of truth for where the error message in English exists. I'll update the issue to explore this too!

robin-macpherson: Cool sounds good!! 🙂

backend/src/graphql.ts
robin-macpherson: This doesn't currently work due to `CommentWhereUniqueInput` only accepting `id` currently, so I have left a comment so we can investigate that separately.

Lanny: Hmm, if it doesn't let use filter by author ID when updating, we need to fetch the comment first and check that its authorId matches the userId from the request (or that the userId belongs to a superuser/moderator/someone who should be allowed to update the comment as opposed to just a logged in user)

robin-macpherson: @Lanny cool that's what I tried to do originally before understanding the whole transactions support thingy. Updated this now to be more robust!

robin-macpherson: Updated the checks on `deleteComment`, too.

robin-macpherson: @Lanny also created a `hasPostPermissions` util function that can be used for checking permissions for creating/updating/deleting a `Post` or `Comment` 😄

backend/src/utils.ts
```diff
@@ -28,3 +29,18 @@ const htmlifyEditorNode = (node: NodeType): string => {
...
+// Takes in an original Post or Comment and a currently logged in User and checks that
+// the currentUser has permission to update or delete that Post/Comment
+export const hasPostPermissions = (
+  original: Post | Comment,
+  currentUser: User,
+) => {
+  const hasPermission =
+    original.authorId == currentUser.id ||
+    currentUser.userRole !== 'MODERATOR' ||
+    'ADMIN'
```
Lanny, line 42:
```suggestion
    currentUser.userRole !== 'ADMIN'
```

Think this will always return `true` as it stands.

That's cool that the same thing can be used for both posts and comments though

backend/src/graphql.ts
robin-macpherson: @Lanny I just added some optimizations here with `Promise.all()`

PR #108, opened 2020-06-25
Implement My Feed Search, Filtering, and Pagination - Backend

### LINKED ISSUE #107: Implement My Feed Search, Filtering, and Pagination
Our "My Feed" page needs to be bold and robust with superb searching and filtering capabilities and good pagination.

closes #108 
closes #302 

#### Perceived Problem

- Users can't search or filter!
- Pages can't paginate!

- Build out the above functionality!

robin-macpherson:
## Description

**Issue:** #107 

Our "My Feed" page needs to be bold and robust with superb searching and filtering capabilities and good pagination.
This PR implements this functionality!

Also fixes our backend `.tsconfig.json` file finally :)

## Subtasks

- [x] Implement filters
- [x] Implement string search
- [x] Implement pagination
- [x] Implement language query

### DIFF
--- a/backend/src/graphql.ts
+++ b/backend/src/graphql.ts
@@ -114,7 +114,6 @@ schema.queryType({
       t.list.field('feed', {
         type: 'Post',
         args: {
-          status: stringArg(),
           search: stringArg({ required: false }),
           language: intArg({ required: false }),
           topic: stringArg({ required: false }),
--- a/backend/src/graphql.ts
+++ b/backend/src/graphql.ts
@@ -117,12 +117,56 @@ schema.queryType({
         type: 'Post',
         args: {
           status: stringArg(),
+          search: stringArg({ required: false }),
+          language: intArg({ required: false }),
+          topic: stringArg({ required: false }),
+          skip: intArg(),
+          first: intArg(),
         },
         resolve: async (parent, args, ctx) => {
+          const filterClauses = []
+
+          if (args.language) {
+            filterClauses.push({
+              language: {
+                id: {
+                  equals: args.language,
+                },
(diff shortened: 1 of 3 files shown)


backend/src/graphql.ts
```diff
@@ -117,12 +115,56 @@ schema.queryType({
...
+            filterClauses.push({
+              OR: [
+                {
+                  title: {
+                    contains: args.search,
+                  },
+                },
+                {
+                  body: {
+                    contains: args.search,
```
Lanny, line 155: As we get more posts, `LIKE` queries can be very taxing on the DB without an index, and title and body aren't indexed (I think) so we should add an issue or something to the kanban as a reminder to add indexes to these (probably after we have a path forward on the case sensitivity thing)

backend/src/graphql.ts
```diff
@@ -117,12 +115,56 @@ schema.queryType({
...
+                  },
+                },
+              ],
+            })
+          }
+
           return ctx.db.post.findMany({
             where: {
-              status: args.status as any,
```
Lanny, line 124: just noticed status fell out of the query here. We probably shouldn't take it as an arg and just enforce search to published posts (I assume we only want authors to be able to read draft posts, plenty of other places we need to add this check)

backend/src/graphql.ts
robin-macpherson: Issue created! See #111

backend/src/graphql.ts
robin-macpherson: Ok removed from args and we can chat about other use cases later

PR #110, opened 2020-06-25
 Add mutations for adding native/learning languages.

Lanny:
## Description

Is necessary for a language select on post creation, probably useful in other places as well


backend/src/graphql.ts
```diff
@@ -166,6 +166,12 @@ schema.queryType({
           return ctx.db.user.findMany()
         },
       }),
+      t.list.field('languages', {
+        type: 'Language',
+        resolve: async (parent, args, ctx) => {
+          return ctx.db.language.findMany()
+        },
+      }),
```
robin-macpherson, line 174: I've added a `languages` query in #108 that allows us to fetch languages that have published posts to populate the filter dropdown, but of course you need the complete list of available languages here. I'm wondering I should just make this query more flexible when this gets merged into master and then my PR to accommodate both use cases, or whether we need two separate queries, `postLanguages` and `feedLanguages` or something like that

Lanny: yup, one query with optional filtering makes more sense, removing the languages query from this PR.

PR #114, opened 2020-06-27
A few things around post creation

Lanny:
## Description

- Generate excerpt when a post is created
- FE and BE for setting the language a post is in
- Made new post page less hideous

Also fixed issue were we had to refetch the currentUser query (or any query relying on cookies) in a `useEffect`

### DIFF
--- a/backend/src/utils.ts
+++ b/backend/src/utils.ts
@@ -7,12 +7,23 @@ type NodeType = {
 }
 
 const typeToElStrMap = {
-  'block-quote': 'blockquote',
+  'heading-one': 'h1',
   'heading-two': 'h2',
+  'block-quote': 'blockquote',
   'bulleted-list': 'ul',
   'numbered-list': 'ol',
-  'paragraph': 'p',
+  paragraph: 'p',
 }
+
+const nonBodyTypes = new Set(['heading-one', 'heading-two', 'block-quote'])
+
+const breakCharacters = new Set([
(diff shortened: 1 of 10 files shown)


backend/src/utils.ts
```diff
@@ -8,13 +8,24 @@ type NodeType = {
...
+const nonBodyTypes = new Set(['heading-one', 'heading-two', 'block-quote'])
+
+const breakCharacters = new Set([
+  ' ', // Good ole ASCII space
+  '.',
+  ';',
+  '　', // Full width space (Japanese, Chinese?)
+  '。', // Ideographic full stop
+])
+
```
robin-macpherson, line 28: Very noiceeeee, thanks for adding this in!! Since I type almost every day in Chinese or Japanese, I deal with the full-width spaces and 。ideographic full stop all the time. Glad we have this covered!

frontend/pages/dashboard/new-post.tsx
```diff
@@ -18,38 +19,65 @@ const initialValue = [
...
+        </div>
+
         <div className="button-container">
-          <Button type="submit" variant={ButtonVariant.Primary}>
+          <Button
+            type="submit"
+            onClick={createNewPost}
+            disabled={!title || langId === -1}
+            variant={ButtonVariant.Primary}
+          >
```
robin-macpherson, line 80: Ooooh thanks for adding in the `disabled` bits!

PR #116, opened 2020-06-29
Robin/make signup n login lavly

### LINKED ISSUE #115: Signup & Login Form Are Totally Out Of Date 😬
#### Perceived Problem

- Signup & Login Forms haven't been updated fully and work a wee bit hideously 😬
- Signup form doesn't re-direct the user to their lavly dashboard

- Handle both the above with class and style!

robin-macpherson:
## Description

**Issue:** closes #115 

A tiny-but-mighty, much needed wee PR to update our Signup and Login experiences to have our brand-spanking new form validation stuff, add our design system stuff to the form headers (and also center them 👌🏼), and re-drect the user to their lavly dashboard after they sign up! 💅🏼

Check out the splendid film below to see the entire thang in action!

**UPDATE:** Decided to also fix the **landing page** and **about page** headers while I'm here! Sadly, these updates are not included in the film.
**UPDATE TWO:** Threw in fixes to blog page as well. Considering shooting a new film... 🎬
**UPDATE THREE:** Okay at this point let's just fix up that bloody new post page while we're at it (see screenshot)!
**UPDATE FOUR:** Fixed routine issue when clicking on `PostCard`s from `MyFeed`

## Subtasks

- [x] Update both forms to use our new form thingy
- [x] Handle re-direct 👉🏼
- [x] Add some nicer styling to our validation messages and form field
- [x] Fix headers on the landing page
- [x] Fix Blog page background color
- [x] Make new post page look all gawjus! 💃🏻
- [x] Fix routine issue with `PostCard`s
- [x] Fix Header on `My Feed` page

## A Gorgeous "Feature" Film
form-improvements

## The New, "New Post" Page

image

### DIFF
--- a/frontend/components/LoginForm/LoginForm.tsx
+++ b/frontend/components/LoginForm/LoginForm.tsx
@@ -167,6 +167,7 @@ const LoginForm: React.FC = () => {
 
         :global(label > p) {
           color: ${theme.colors.red};
+          font-style: italic;
         }
         
         :global(.form-error) {
--- a/frontend/components/LoginForm/LoginForm.tsx
+++ b/frontend/components/LoginForm/LoginForm.tsx
@@ -17,7 +17,14 @@ const LoginForm: React.FC = () => {
 
   const fieldErrorName = Object.keys(errors)[0] || ''
 
-  const [loginUser, { loading, error }] = useLoginUserMutation()
+  const [loginUser, { loading, error }] = useLoginUserMutation({
+    onCompleted: () => {
+      trackLogIn()
+      router.push({
+        pathname: '/dashboard/my-feed',
+      })
+    },
+  })
 
   const onSubmit = (data: any) => {
     if (!loading && Object.keys(errors).length === 0) {
@@ -27,10 +34,6 @@ const LoginForm: React.FC = () => {
           password: data.password,
         },
         refetchQueries: [{ query: CurrentUserDocument }],
-      })
-      trackLogIn()
-      router.push({
-        pathname: '/dashboard/my-feed',
       })
     }
   }
--- a/frontend/components/LoginForm/LoginForm.tsx
+++ b/frontend/components/LoginForm/LoginForm.tsx
@@ -57,7 +57,7 @@ const LoginForm: React.FC = () => {
               },
             })}
           />
-          <ErrorMessage errors={errors} name="email" as="em" />
+          <ErrorMessage errors={errors} name="email" as="p" />
         </label>
         <label htmlFor="password">
           Password
@@ -70,7 +70,7 @@ const LoginForm: React.FC = () => {
               minLength: { value: 6, message: 'Password must be at least 6 characters' },
             })}
           />
-          <ErrorMessage errors={errors} name="password" as="em" />
+          <ErrorMessage errors={errors} name="password" as="p" />
         </label>
 
         <Button type="submit">Log In</Button>
@@ -165,7 +165,7 @@ const LoginForm: React.FC = () => {
           box-shadow: 0px 8px 10px #00000029;
         }
 
-        :global(label > em) {
+        :global(label > p) {
           color: ${theme.colors.red};
         }
         
--- a/frontend/components/LoginForm/LoginForm.tsx
+++ b/frontend/components/LoginForm/LoginForm.tsx
@@ -1,3 +1,4 @@
+import React from 'react'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useForm, ErrorMessage } from 'react-hook-form'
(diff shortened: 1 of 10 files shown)


frontend/components/LoginForm/LoginForm.tsx
```diff
@@ -50,7 +54,7 @@ const LoginForm: React.FC = () => {
               },
             })}
           />
-          <ErrorMessage errors={errors} name="email" as="p" />
+          <ErrorMessage errors={errors} name="email" as="em" />
```
nlitwin, line 57: Semantically, we'd want to keep the `p` tag and use `font-style: italic`

frontend/components/SignupForm/SignupForm.tsx
```diff
@@ -1,98 +1,90 @@
...
-          handle: values.handle,
-          email: values.email,
-          password: values.password,
+          handle: data.handle,
+          email: data.email,
+          password: data.password,
         },
       })
+      trackCreateAccount()
+      router.push({
```
nlitwin, line 32: Navigating to the next page should stay in the `onCompleted` callback of the `createUser` mutation in case there is an error. I'm also thinking that the `trackCreateAccount` should go there too, since it would look like the user has been created from the event, but an error would prevent the user from being created.

frontend/components/SignupForm/SignupForm.tsx
```diff
@@ -1,98 +1,90 @@
...
   return (
-    <form onSubmit={handleSubmit}>
+    <form onSubmit={handleSubmit(onSubmit)}>
       <fieldset disabled={loading} aria-busy={loading}>
         <h2>Sign up for an account</h2>
         {error && <FormError error={error} />}
         <label htmlFor="handle">
-          Username
-          <p>{errors?.handle}</p>
+          Display Name
```
nlitwin, line 44: Can we translate all text that appears in the form? `/frontend/public/static/locales/en/settings.json` has translations for all the settings form values, like labels and form error messages, if you want to use that as a guide for structuring a new translation file. We could create something like `account.json` perhaps that could house text for both the login and signup forms (I'm assuming some text values will be the same across the two forms)

frontend/components/SignupForm/SignupForm.tsx
```diff
@@ -1,98 +1,90 @@
...
             name="handle"
-            value={values.handle}
-            placeholder="Your username"
+            placeholder="Your name, or perhaps a fun pseudonym!"
             autoComplete="on"
-            onChange={handleChange}
-            onBlur={handleBlur}
+            ref={register({
+              required: 'Display Name is required',
+              minLength: { value: 6, message: 'Your display name must be at least 6 characters' },
```
nlitwin, line 52: 😱 I wonder if there's a way for me to have a special, shorter display name - just "Nick" 🤔 🤣

robin-macpherson: lol for now I just put 6 chars but we can chat about what it should be 😉

frontend/components/SignupForm/SignupForm.tsx
robin-macpherson: As discussed in my audio message, for this PR I was just trying to make a bunch of these quick fixes for things that are really broken aesthetically or that were major blockers to the UX or even completing a flow (like the routing issue). I propose to handle this in another PR 🙂

nlitwin: Sounds good!!

frontend/components/LoginForm/LoginForm.tsx
robin-macpherson: Isn't using `em` and `strong` considered to be the correct approach for semantic markup and also for screen readers?

frontend/components/SignupForm/SignupForm.tsx
robin-macpherson: Good call, I'll move it in the `LoginForm` as well then 👍🏼

robin-macpherson: Done like dinner! 🍗

frontend/components/SignupForm/SignupForm.tsx
robin-macpherson: @nlitwin made it 3 chars for ya! That feels reasonable for a display name 😌

frontend/components/LoginForm/LoginForm.tsx
nlitwin: In this case it's more of a design effect and "style" than actually emphasizing text. `em` and `strong` as semantic elements are generally reserved for linguistic usage. From the [HTML spec doc](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-em-element):

> The placement of stress emphasis changes the meaning of the sentence. The element thus forms an integral part of the content. The precise way in which stress is used in this way depends on the language.

frontend/components/SignupForm/SignupForm.tsx
nlitwin: Woooo hoooo!!

frontend/components/SignupForm/SignupForm.tsx
nlitwin: Done and dusted 🧹 ✨

frontend/components/LoginForm/LoginForm.tsx
robin-macpherson: Okie dokie changed 😄

Issue #112, opened 2020-06-27
Add `currentUser` to application state via Apollo


robin-macpherson:
#### Perceived Problem

- Currently, in order to get the `currentUser` we have to call that query in whichever component needs it.
- It is currently being fetched in the Nav, but we don't even have the ability to pass that to other pages.

#### Ideas / Proposed Solution(s)

- Build out our application state to make this available via Apollo

nlitwin:
Looks like I'll need this for the profile page, so I'll grab it :)

robin-macpherson:
> Looks like I'll need this for the profile page, so I'll grab it :)

Absolutely smashing kind sir!

PR #117, opened 2020-06-29
Write and read current user data

nlitwin:
## Description

**Issue:** #112

Multiple components in the dashboard need access to the currently logged in user.

🚨 Note: I'm currently getting this error:

@robin-macpherson @Lanny Feel free to take this PR over if you want, otherwise I can continue looking at it later today 🙂 

## Subtasks

- [ ] ...

### DIFF
--- a/frontend/components/Dashboard/Nav/Nav.tsx
+++ b/frontend/components/Dashboard/Nav/Nav.tsx
@@ -3,7 +3,8 @@ import Link from 'next/link'
 import { navConstants } from './nav-constants'
 import NavLinks from './NavLinks'
 import { black } from '../../../utils'
-import { useCurrentUserQuery } from '../../../generated/graphql'
+import { CurrentUserDocument } from '../../../generated/graphql'
+import { useApolloClient } from '@apollo/client'
 
 interface Props {
   expanded: boolean
@@ -11,13 +12,13 @@ interface Props {
 }
 
 const Nav: React.FC<Props> = ({ expanded, collapse }) => {
-  const { data, refetch } = useCurrentUserQuery()
+  const client = useApolloClient()
+  const user = client.readQuery({ query: CurrentUserDocument })
 
   useEffect(() => {
     setTimeout(() => {
       document.body.classList.remove('block-transitions-on-page-load')
     }, 0)
-    refetch()
   }, [])
 
   const handleCollapse = (): void => {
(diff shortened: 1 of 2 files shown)


nlitwin:
In case it's helpful for debugging, I found that if you wrap the query in `Nav.tsx` in a try catch, it will allow you to render the pages:

```
const user = client.readQuery({ query: CurrentUserDocument })
  let user
  try {
    user = client.readQuery({ query: CurrentUserDocument })
  } catch (e) {
    user = null
  }
```

I found that if you go to `dashboard/login`, login as a user and then get redirected to the dashboard, the `currentUser` query from the cache works fine, but if you just visit `/dashboard/my-feed` directly without coming from `/login`, then the `ROOT_QUERY` error appears. It seems like somewhere we need to add a default value to the cache, perhaps setting `currentUser` to null?

Lanny:
Unsure how related it is, but I think `useApolloClient` will skip our client setup in lib/apollo, which I recently made some changes to to support SSR fetching of data that uses cookies (currentUser) here:

https://github.com/Journaly/journaly/pull/114/files#diff-773c7e2499be935f036213f03a318d18R58

Lanny:
oh, I see, readQuery won't attempt to fetch if there's nothing in the store or cache or whatever and on a fresh page load there's nothing there. Won't the normal `useCurrentUserQuery` consult the cache before fetching? Why change it to readQuery here?

nlitwin:
@robin-macpherson Correct me if you had a different idea, but I think what we want is for any component of the app to immediately know if there is a `currentUser` or not. So if there isn't, the cache will just return `null` or an empty object and the component can render immediately. Right now `useCurrentUserQuery` has to check and perhaps make an API call introducing a `loading` state to each component that would need access to it.

If there is a way for us to initialize the cache with an empty value and then add a `currentUser` value to the cache once a user has authenticated via login/signup, that would probably be ideal.

Hmm good point about `useApolloClient`. I wonder what the right way is to get access to the client and/or cache?

robin-macpherson:
@nlitwin so @Lanny and I looked into this and his suspicion was correct -- Apollo is already handling caching the `currentUser` for us with our existing set up, so we can actually close this and continue fetching the `currentUser` with the `useCurrentUserQuery()` in these components where needed, and the request will not actually be made if the data is already in the cache.

Issue #76, opened 2020-05-31
Prepare Cypress for first tests


nlitwin:
#### Perceived Problem

- Our Cypress infra needs to be fleshed out a bit more in preparation for e2e tests

#### Ideas / Proposed Solution(s)

- Wire things up to prepare for cypress tests, e.g. add more to docs, update Cypress version, add some examples

PR #118, opened 2020-06-30
Add First Journaly Cypress Integration Tests

robin-macpherson:
## Description

**Issue:** #76 

This PR sets up and creates our first Integration tests for Cypress.

- Adds Cypress to the root of our project so that it can run integration tests on the entire app and later run the app.
- Sets up basic test organization closely following the exact recommendations from the Cypress founder.
- Adds a custom command for login while also providing a best practice pattern to strive towards but needs more investigation due to GraphQL.
- Adds first integration tests

**NOTE:** *The intent of this PR is to keep it fairly light and just get Cypress set up for Integration Tests, introduce general organization, provide some patterns and tests as references, while also writing our first tests. However, the intent isn't for the tests to be 100% perfect and fully-fledged just yet. I will submit subsequent PRs that add documentation and continue add more tests in batches.*

Please also keep in mind that we are using Cypress for unit tests as well, and so that is why we also have Cypress as a dependency in the `frontend` & `backend` directories.

## Subtasks

- [x] Set up Cypress in project root
- [x] Basic configuration (`baseUrl`)
- [x] Add first tests
- [x] Add test script
- [x] Test... well, the tests!

### Login Test

cypress-login-test-demo

### Signup Test

cypress-signup-test-demo

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,1 @@
+{ "baseUrl": "http://localhost:3000" }
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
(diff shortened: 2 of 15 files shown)


robin-macpherson:
> c'est parfait! 🐉 🗻

Yay ☀️🎈🦔💨

PR #124, opened 2020-07-01
Add The Almighty `AuthGate`, AKA "Gatekeeper Of Auth" 🤺⾨

### LINKED ISSUE #123: Create An Authentication Gate 🤺⾨
#### Perceived Problem

- Currently, auth-protected pages are accessible and we don't have a good way to protect them.

- Create a super nice, clean, and easy-to-use component that handles auth checking while taking advantage of our `LoadingWrapper` Component.

robin-macpherson:
## Description

**Issue:** closes #123 

Currently, auth-protected pages are accessible and we don't have a good way to protect them.

### Wha... What's That Sound!?

... *Sound of a ninja swinging in on a rope* ...
😳💥🤺💨 ... ⾨  - **"Haaaaaaa-ya!!!"**

This PR creates a super nice, clean, and easy-to-use component, `<AuthGate>` AKA "AuthGatekeeper" AKA "Gatekeeper Of Auth" that handles authentication checking while taking advantage of our existing `LoadingWrapper` Component.

This PR also handles clearing the `currentUser` in the Apollo Cache upon logging in and signing up to make sure it gets updated with the newly authenticated user and they are redirected properly.

## Subtasks

- [x] Create `AuthGate`
- [x] Handle login and signup cache things
- [x] Put up the gate around protected pages!

## A Short Film You'll Love

auth-gate-demo


frontend/components/LoginForm/LoginForm.tsx
```diff
@@ -3,7 +3,11 @@ import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useForm, ErrorMessage } from 'react-hook-form'
 import { trackLogIn } from '../../events/users'
-import { useLoginUserMutation, CurrentUserDocument } from '../../generated/graphql'
+import {
+  useLoginUserMutation,
+  CurrentUserDocument,
```
Lanny, line 8: I don't think this ended up being used anywhere, did it?

frontend/components/AuthGate/AuthGate.tsx
```diff
@@ -0,0 +1,29 @@
...
+import LoadingWrapper from '../LoadingWrapper'
+import { useCurrentUserQuery } from '../../generated/graphql'
+
+/**
+ * Checks that the user is logged in.
+ * If not, redirects them to the login page.
+ */
+
+type AuthGateProps = {
+  children: ReactElement
```
nlitwin, line 12: nit: it's usually best to default to the `ReactNode` type since it encompasses more possible values that we'd want to use potentially:

```js
type ReactChild = ReactElement | ReactText;
type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

nlitwin, line 12: nit: it's usually best to default to the `ReactNode` type since it encompasses more possible values that we'd want to use potentially:

```js
type ReactChild = ReactElement | ReactText;
type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

frontend/pages/dashboard/my-feed.tsx
```diff
@@ -3,28 +3,41 @@ import { NextPage } from 'next'
...
 interface InitialProps {
   namespacesRequired: string[]
 }
 
 const MyFeedPage: NextPage<InitialProps> = () => {
   const { t } = useTranslation()
-  const { loading, error, data } = useFeedQuery({
+  const {
+    data: currentUserData,
+    loading: currentUserLoading,
```
nlitwin, line 19: Maybe I don't yet quite understand the structure, but I'm not sure why we should need to have a `useCurrentUserQuery` in any of the pages if `AuthGate` already makes that API call and handles loading/errors. I think what would happen in this file is that `useCurrentUserQuery` would be triggered, but `AuthGate` would immediately render and also make that call.

If we make `AuthGate` a render prop component, which exposes `currentUser` to the children, then the pages could immediately grab `currentUser` and not have to handle `loading` and `error` states in addition to any other `useXQuery`'s `loading` and `error` states. In the render prop case, a logged out user would just immediately get redirected, and for a logged in user, the cached value would be immediately available to the children.

```jsx
<AuthGate>
  {(currentUser) => (
    <DashboardLayout>
      <div>{currentUser.name} is the best}</div>
    </DashboardLayout>
  )
</AuthGate>
```

Let me know if that makes sense! 🙂

frontend/components/LoginForm/LoginForm.tsx
robin-macpherson: `CurrentUserDocument`? Yep it's used on line 43 in the following code block:

```ts
const onSubmit = (data: any) => {
    if (!loading && Object.keys(errors).length === 0) {
      loginUser({
        variables: {
          identifier: data.email,
          password: data.password,
        },
        refetchQueries: [{ query: CurrentUserDocument }],
      })
    }
  }
```

frontend/components/AuthGate/AuthGate.tsx
Lanny: VSC yelled when we tried that, but maybe because LoadingWrapper also expects ReactElement, if we can satisfy TSC with ReactNode then that's good with me

frontend/pages/dashboard/my-feed.tsx
Lanny: Yeah, passing down currentUser makes sense to me 👍 . I don't think there's much of a perf difference since one one fetch should ever go through, but not having to handle loading/error states would be a win.

robin-macpherson: Yeah @nlitwin as mentioned in the other PR that we closed earlier, we tested that no further calls are made with queries that have already been cached which navigating between pages in the app, so the additional calls would not be made. But yep I can work on simplifying it and passing down `currentUser`.

frontend/components/AuthGate/AuthGate.tsx
robin-macpherson: Yeah should be as simple as just updating the type definition in `LoadingWrapper`. We had it as `ReactNode` first but this was just an effort to avoid updating too much stuff, I can go and update that and it should work

robin-macpherson: Actually, changing it to `ReactNode` in `LoadingWrapper` does indeed seem to break things and cause VS Code to shout. We can look into that more later if we want but I would advocate for not holding this PR up over that now.

frontend/pages/dashboard/my-feed.tsx
robin-macpherson: Been having issues getting this all to work as requested as well, so unfortunately this is going to get held up another day now. Will get back to it tomorrow

nlitwin: Sounds good 👍

PR #127, opened 2020-07-02
Expose currentUser directly to children of AuthGate

### LINKED ISSUE #112: Add `currentUser` to application state via Apollo

nlitwin:
## Description

**Issue:** #123, closes #112

This PR follows up on #123 and PR #124 implementing direct access to the `user` as discussed [here](https://github.com/Journaly/journaly/pull/124#discussion_r448082301)

## Subtasks

- [x] Give children components of `AuthGate` direct access to the `user` without having to deal with `loading` or `error` states
  - [x] Allow non-function children too for cases when your page is wrapped in `AuthGate` but you don't need access to any properties on `user`

  ```jsx
    /* allow both versions */
    <AuthGate>
      <Child />
    </AuthGate>

    <AuthGate>
      {(currentUser) => (
        <Child currentUser={currentUser} />
      )}
    </AuthGate>

  ```

## Screenshots

No new changes visually


frontend/components/Dashboard/Dashboard.tsx
```diff
@@ -3,12 +3,14 @@ import { useRouter } from 'next/router'
 import Nav, { navConstants } from './Nav'
 import Header from './Header'
 import useWindowSize from '../../hooks/useWindowSize'
+import { User } from '../../generated/graphql'
```
robin-macpherson, line 6:
```suggestion
import { User as UserType } from '../../generated/graphql'
```

robin-macpherson: ^ just in accordance with the convention we've been following everywhere else

frontend/components/Dashboard/Dashboard.tsx
```diff
@@ -3,12 +3,14 @@ import { useRouter } from 'next/router'
...
 
 interface Props {
+  user?: User
```
robin-macpherson, line 9:
```suggestion
  user?: UserType
```

frontend/components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -2,17 +2,17 @@ import React from 'react'
 
 import PostCard from '../PostCard'
 import theme from '../../../theme'
-import { User as UserType } from '../../../generated/graphql'
+import { User } from '../../../generated/graphql'
```
robin-macpherson, line 5:
```suggestion
import { User as UserType } from '../../../generated/graphql'
```

frontend/components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -2,17 +2,17 @@ import React from 'react'
...
 
 // import { Post } from '../../../generated/graphql' // TODO (robin-macpherson): See Post[] type todo below
 
 type Props = {
   posts: any[] // TODO (robin-macpherson): should be Post[]
-  currentUser: UserType
+  user: User
```
robin-macpherson, line 11:
```suggestion
  user: UserType
```

frontend/components/Dashboard/Nav/Nav.tsx
```diff
@@ -3,16 +3,15 @@ import Link from 'next/link'
 import { navConstants } from './nav-constants'
 import NavLinks from './NavLinks'
 import { black } from '../../../utils'
-import { useCurrentUserQuery } from '../../../generated/graphql'
+import { User } from '../../../generated/graphql'
```
robin-macpherson, line 6:
```suggestion
import { User as UserType } from '../../../generated/graphql'
```

frontend/components/Dashboard/Nav/Nav.tsx
```diff
@@ -3,16 +3,15 @@ import Link from 'next/link'
...
 
 interface Props {
+  user?: User
```
robin-macpherson, line 9:
```suggestion
  user?: UserType
```

frontend/components/Dashboard/Nav/NavLinks.tsx
```diff
@@ -5,21 +5,17 @@ import FeedIcon from '../../Icons/FeedIcon'
...
 
 interface Props {
   onClick: () => void
-  // TOOD (robin-macpherson): solve issue with `UserType` requiring `posts`
-  currentUser: UserType | any
+  user: User
 }
 
-const NavLinks: React.FC<Props> = ({ onClick, currentUser }) => {
+const NavLinks: React.FC<Props> = ({ onClick, user }) => {
```
robin-macpherson, line 15: I would like us to stick with `currentUser`, as it is more descriptive as that was the reason for that name since the beginning.

frontend/components/Dashboard/Dashboard.tsx
nlitwin: I actually was under the impression that our convention was using the name as it is generated from the graphql file and only adding the `as XType` alias when it conflicted with another variable name. I can change it back

frontend/components/Dashboard/Nav/NavLinks.tsx
nlitwin: I'd like to gently push back on this one, since it will be present throughout the app forever :D For the most part, our app will only really have one "user" and it is much less verbose than having to write `currentUser={currentUser}` as a prop and value to any and all components that need it. In all other cases, the other users that appear in the app are `author`s, and even if we have a search page for users some day, we can always name a local variable `currentUser` should the need for clearing up ambiguity arise

robin-macpherson: I don't agree. I think having good, descriptive names for things like this is well worth a few extra characters. We're going to have tons and tons of events firing all over the place, and if you think more in terms of how our data is structured and represented in the database, everyone is a `User` and they all ultimately live in the `User` table, even if there is a join on `authorId`. I would prefer that we think carefully in terms of how our data is structured and represented in the underlying database and implications with things like running data analysis on huge data sets and so on. And as you mentioned, there will be situations like search, as well as later things like instant messaging, friendships, accountability buddies, and many other cases where `User`s will be represented in various ways. Even the idea that well we'll likely have those situations one day but then we can just give local variables different names, etc. feels wrong to me.

robin-macpherson: Not to mention if we consider some of the workflow benefits of having all this TypeScript set up with the superb auto-completion, you'll likely type the same number of chars for both anyway if you're utilizing that tooling. I still think it would be totally worth being declarative and explicit regardless, but just saying that in terms of extra keystrokes, our setup makes that a bit of a moot point if you're using VS Code and all that tooling, which you are.

robin-macpherson: Lastly, if we continue thinking about characters and syntax, if your next argument would be that we could simply rename certain values when logging events and so on, we'd end up typing way more characters since we wouldn't be able to take advantage of nice, clean object property syntax and have to specify each one. Ideally I'd like to not have to do lots of data cleaning behind the scenes for our data science initiatives.

robin-macpherson: Again, thinking about this as narrowly as us just having one "user" and everyone else is an "author" I think fails to consider to rich feature roadmap that we have ahead of us and sets us up for having to either change all this stuff later, or lead to inconsistent naming all over the place down the line.

robin-macpherson: @nlitwin one last thing I'll say on this is that even if we wanted to discuss this, I don't think it should have been in this PR. This was supposed to be a super quick PR following #124 to directly expose the `currentUser` as an improvement on that pattern that would get in quickly. That was the whole reason why we decided to merge #124 as it was, because we're so short on time. I believe this PR should have stuck to that focus, rather than trying to also make changes to the way we name and represent/think about the `currentUser` and then having this sit for potentially another day.

nlitwin: cool, happy to change it back 😊 let me know if the recent commits look good to you!

PR #128, opened 2020-07-02
Autosave editor content on new post page.

Lanny:
Title says it all.

Do we want any kind of visual indication that a post is saved locally? It wouldn't be too hard to add a little piece of copy that says like "unsaved changes" or "saved locally" or something. I didn't add it because my thought was that you don't really care about the feature until you fat finger something and think you've lost your post, at which point you'll discovery it was saved on going back, but if you think it will make folks more comfortable to have feedback on it right away we can throw that in.

### DIFF
--- a/frontend/hooks/useAutosavedState.ts
+++ b/frontend/hooks/useAutosavedState.ts
@@ -1,15 +1,10 @@
 import React from 'react'
 
-type optionsType = {
-  key: Maybe<string>
-  debounceTime: Maybe<number>
-}
-
 export default function useAutosavedState<T>(
   initialValue: T,
   opts = { key: 'default', debounceTime: 1000 },
-) {
-  const storage = (typeof window !== 'undefined' && window.localStorage) || {}
+): [T, (value: T) => void, () => void] {
+  const storage: any = (typeof window !== 'undefined' && window.localStorage) || {}
   const storageKey = `autsave[${opts.key || 'default'}]`
 
   const [value, setValue] = React.useState<T>(initialValue)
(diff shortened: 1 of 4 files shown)


frontend/hooks/useAutosavedState.ts
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+
+export default function useAutosavedState<T>(
+  initialValue: T,
+  opts = { key: 'default', debounceTime: 1000 },
+): [T, (value: T) => void, () => void] {
+  const storage: any = (typeof window !== 'undefined' && window.localStorage) || {}
```
nlitwin, line 7: Could there be a mismatch where SSR starts out with the empty object literal, and then `localStorage` is used on the client?

Lanny: This should get re-evaluated on every render, so if it gets SSR'd, then handed off to the browser, it should just switch to using local storage. Reloading the page on browser (so doing an SSR) it seemed to hang onto my content without any issue.

robin-macpherson:
Looks great! Regarding the question in the PR description, I think it would be nice to have some visual treatment for indicating that we're doing this nice thing for them, however it could be misleading if we aren't actually auto-saving for real and it's just a `localStorage` safety net for fat-finger stuff. I still want to think about the UX decision behind real auto-saving or not, so for now I say let's merge this as is and have a follow-up if need be 😄

frontend/hooks/useAutosavedState.ts
Lanny: Oh, and that behavior would be desirable. Since there's not really anything to be done here during SSR (we don't have this content on the server) just using an empty object and not restoring anything is what we want.

PR #134, opened 2020-07-03
Display published and draft posts on the My Posts page

### LINKED ISSUE #103: My Posts and Drafts Page
#### Perceived Problem

- Users will need to have a place to view all of their posts as well as their drafts.

- Create a single page for viewing my posts and drafts with a toggle to switch between the two

Drafts wireframe (My Posts will sit on the same page and will be accessible via a central toggle button)

WhatsApp Image 2020-06-07 at 12 02 14 PM (4)

nlitwin:
## Description

**Issue:** Closes #103 

Now that we have a TabToggle on the My Posts page, we can add published and draft posts 

## Subtasks

- [x] Support querying for posts based on status

### Desktop (Draft)

localhost_3000_dashboard_my-posts (3)

### DIFF
--- a/backend/src/graphql.ts
+++ b/backend/src/graphql.ts
@@ -150,12 +150,16 @@ schema.queryType({
     t.list.field('posts', {
       type: 'Post',
       args: {
+        status: arg({ type: 'PostStatus', required: false }),
         authorId: intArg({ required: true }),
       },
       resolve: async (_root, args, ctx) => {
         return ctx.db.post.findMany({
           where: {
-            author: { id: args.authorId },
+            AND: {
+              author: { id: args.authorId },
+              status: args.status || 'PUBLISHED',
+            },
           },
         })
       },
--- a/backend/src/graphql.ts
+++ b/backend/src/graphql.ts
@@ -150,7 +150,7 @@ schema.queryType({
     t.list.field('posts', {
       type: 'Post',
       args: {
-        status: arg({ type: 'PostStatus', required: false }),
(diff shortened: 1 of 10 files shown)


backend/src/graphql.ts
```diff
@@ -150,12 +150,16 @@ schema.queryType({
     t.list.field('posts', {
       type: 'Post',
       args: {
+        status: arg({ type: 'PostStatus', required: false }),
```
robin-macpherson, line 153: `PostStatus` should be `required: true` 👍🏼 (every post should have a status and we should always pass the one we want here)

backend/src/graphql.ts
```diff
@@ -150,12 +150,16 @@ schema.queryType({
...
         authorId: intArg({ required: true }),
       },
       resolve: async (_root, args, ctx) => {
         return ctx.db.post.findMany({
           where: {
-            author: { id: args.authorId },
+            AND: {
+              author: { id: args.authorId },
+              status: args.status || 'PUBLISHED',
```
robin-macpherson, line 161: This should be just `args.status`

frontend/graphql/posts.graphql
```diff
@@ -1,5 +1,5 @@
-query posts($authorId: Int!) {
-  posts(authorId: $authorId) {
+query posts($authorId: Int!, $status: PostStatus) {
```
robin-macpherson, line 1: We should also make the `$status` arg required here in accordance with the backend request above

nlitwin:
Closed in favor of #140

Issue #79, opened 2020-05-31
Flesh out the "My Feed" page


nlitwin:
#### Perceived Problem

- The My Feed page still needs some more implementation, as the form fields aren't wired up yet

#### Ideas / Proposed Solution(s)

- Wire up the form fields and update the CSS/layout according to the designs

nlitwin:
@robin-macpherson do you happen to have a link to the designs or wireframes for the My Feed page? I'm pretty sure I've seen them at one point, but I can't find a link in Adobe. Thanks!

nlitwin:
![WhatsApp Image 2020-06-07 at 12 02 14 PM](https://user-images.githubusercontent.com/5829188/85171197-636e6480-b223-11ea-9ef4-e072adc97408.jpeg)

nlitwin:
Closing in favor of #107 since this issue is somewhat of a duplicate. Using the `PostCard` component on the My Feed page done and the PR for adding pagination is #166

PR #135, opened 2020-07-03
Use PostCards on the My Feed Page

nlitwin:
## Description

**Issue:** #79 

This PR adds support for displaying avatars in the `<PostCard />` component. It also adds a `stacked` prop, which keeps the card in the stacked arrangement (image on top of post details), even on large screens.

🚨 Note: I wasn't 100% sure how to handle long usernames since they go up against the "stats" (number of likes and comments). We may need to stack the avatar above the stats or change the arrangement a little, but I thought I'd get the basic version implemented that was in the wireframes (see image in #79). I did notice that the wireframes wrap the name onto two lines, but wrapping wouldn't quite work for users' names consisting of one solid long word.

## Subtasks

- [x] Add support for stacked `PostCard`s
- [x] Add support for displaying avatars / author name in `PostCard`s
- [x] Make sure non-stacked `PostCard`s still look good

## Screenshots

image


nlitwin:
Closed in favor of #141

PR #143, opened 2020-07-04
Fix bugs with the side nav

### LINKED ISSUE #144: Fix bugs in the nav
#### Description

The side nav in the dashboard has a couple issues with it at the moment. The nav links are not centered vertically on the page, and the nav itself jumps around when visiting different pages. In addition the nav links are flickering and the background color needs to be updated

#### Repro Steps

1. Visit different pages and shrink/expand the size of the screen

nlitwin:
## Description

**Issue:** Closes #144 

This fixes the nav from jumping around on each page load and also centers the links vertically on the page. It also updates the nav background and hover color

## Subtasks

- [x] Center links vertically on the page
- [x] Fix jumping around when navigating to different pages
- [x] Update the nav background and hover color
- [x] Remove animations on individual links to get rid of "flickering" on page transition

### DIFF
--- a/frontend/components/Dashboard/Nav/Nav.tsx
+++ b/frontend/components/Dashboard/Nav/Nav.tsx
@@ -2,7 +2,6 @@ import React, { useEffect } from 'react'
 import Link from 'next/link'
 import { navConstants } from './nav-constants'
 import NavLinks from './NavLinks'
-import { black } from '../../../utils'
 import { User as UserType } from '../../../generated/graphql'
 
 interface Props {
@@ -48,7 +47,7 @@ const Nav: React.FC<Props> = ({ currentUser, expanded, collapse }) => {
           width: 100%;
           background-color: rgba(0, 0, 0, 0.4);
           opacity: 0;
-          transition: opacity ${navConstants.transitionDuration}ms ease-in-out;
+          transition: opacity ${navConstants.transitionDuration}ms linear;
           z-index: ${navConstants.zIndex};
         }
 
@@ -69,15 +68,15 @@ const Nav: React.FC<Props> = ({ currentUser, expanded, collapse }) => {
           position: fixed;
           top: 0;
           left: 0;
-          display: flex;
-          flex-direction: column;
+          display: grid;
+          grid-template-rows: 1fr 2fr 1fr;
           height: 100vh;
           width: ${navConstants.navWidth}px;
-          background-color: ${black};
+          background-color: #313131;
           z-index: ${navConstants.zIndex};
           transform: translateX(-100%);
-          transition: transform ${navConstants.transitionDuration}ms ease-in-out,
-            width ${navConstants.transitionDuration}ms ease-in-out;
+          transition: transform ${navConstants.transitionDuration}ms linear,
+            width ${navConstants.transitionDuration}ms linear;
         }
 
         .expanded nav {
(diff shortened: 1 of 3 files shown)


frontend/components/Dashboard/Nav/Nav.tsx
```diff
@@ -48,7 +47,7 @@ const Nav: React.FC<Props> = ({ currentUser, expanded, collapse }) => {
           width: 100%;
           background-color: rgba(0, 0, 0, 0.4);
           opacity: 0;
-          transition: opacity ${navConstants.transitionDuration}ms ease-in-out;
+          transition: opacity ${navConstants.transitionDuration}ms linear;
```
nlitwin, line 50: Other animations were using `linear`, so making them all the same easing function helps the animations look more in sync. We can play play around with the exact easing function and update all of the locations at once, if need be

frontend/components/Dashboard/Nav/Nav.tsx
```diff
@@ -69,15 +68,15 @@ const Nav: React.FC<Props> = ({ currentUser, expanded, collapse }) => {
           position: fixed;
           top: 0;
           left: 0;
-          display: flex;
-          flex-direction: column;
+          display: grid;
```
nlitwin, line 71: The _only_ way I was able to get the nav links centered on the screen was to switch to `grid`. Apparently you can't do `justify-self` with `flex` when you're using `flex-direction: column` 😞

frontend/components/Dashboard/Nav/NavLinks.tsx
```diff
@@ -153,16 +152,16 @@ const NavLinks: React.FC<Props> = ({ onClick, currentUser }) => {
...
             animation: fadeIn ${navConstants.transitionDuration}ms linear;
           }
         }
 
         .nav-bottom {
           display: flex;
           flex-direction: column;
-          /* Expands the links to push the Journaly "J" logo down to the bottom */
-          flex-grow: 2;
+          /* centers itself in the grid container */
```
nlitwin, line 163: The container for the nav links is centered vertically with grid already, but this makes the links within the container centered in the container

frontend/components/Dashboard/Nav/NavLinks.tsx
```diff
@@ -174,22 +173,13 @@ const NavLinks: React.FC<Props> = ({ onClick, currentUser }) => {
 
         :global(.expanded) .nav-bottom hr {
           width: 180px;
-          margin: 0 auto;
+          margin: 20px auto;
```
nlitwin, line 176: The `<hr />` had no top/bottom padding and looked weird when the link above and below it were `active`. Adding margin also aligns it with the designs more

frontend/components/Layouts/DashboardLayout.tsx
```diff
@@ -16,7 +16,24 @@ interface Props {
 const DashboardLayout: React.FC<Props> = ({ currentUser, children, withPadding = true }) => {
   const { width } = useWindowSize()
   const router = useRouter()
-  const [navExpanded, setNavExpanded] = useState(false)
+  const shouldNavBeExpanded = () => {
```
nlitwin, line 19: This logic used to be only in the `useEffect`, but I extracted it out into its own function so that we can also use it for the initial value of the `navExpanded` `useState`

PR #151, opened 2020-07-05
Dynamically get currentUser from inside <Nav />

### LINKED ISSUE #150: Fix Nav for pages that don't require authentication
#### Description

If you visit a post at `/post/<id>`, you will see the unauthenticated version of the nav, since that page doesn't require authentication to view. However, if you are logged in, you _should_ see the nav links. Adding `useCurrentUserQuery` to the nav fixes this.

#### Repro Steps

1. Visit localhost:3000/dashboard/post/1
2. See unauthenticated version of nav
3. Log in and go back to `/dashboard/post/1`
4. Still see unauthenticated version of nav

nlitwin:
## Description

**Issue:** Closes #150 

Fixes issue described in #150 as well as the mismatch of classNames between server and client. 

## Subtasks

- [x] Add `useCurrentUserQuery` to `Nav.tsx` so it can load a user's information even when on a page that doesn't require authentication
- [x] Fix mismatch between SSR and client hydration with regards to `className` in the Nav

### Unauthenticated view of the post page

🚨 Note: position of J is fixed in #147 🚨

### DIFF
--- a/frontend/pages/dashboard/new-post.tsx
+++ b/frontend/pages/dashboard/new-post.tsx
@@ -59,93 +59,91 @@ const NewPostPage: NextPage = () => {
 
   return (
     <AuthGate>
-      {(currentUser) => (
-        <DashboardLayout currentUser={currentUser}>
-          <form id="new-post">
-            <h1>Let's write a post</h1>
+      <DashboardLayout>
+        <form id="new-post">
+          <h1>Let's write a post</h1>
 
-            <label htmlFor="post-title">Title</label>
-            <input
-              className="j-field"
-              id="post-title"
-              value={title}
-              onChange={(e) => setTitle(e.target.value)}
-              type="text"
-              name="title"
-              placeholder="The Greatest Story Never Told..."
-              autoComplete="off"
-            />
+          <label htmlFor="post-title">Title</label>
+          <input
+            className="j-field"
+            id="post-title"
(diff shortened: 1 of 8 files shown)


frontend/pages/dashboard/new-post.tsx
```diff
@@ -59,93 +59,91 @@ const NewPostPage: NextPage = () => {
...
-                onClick={(e: React.MouseEvent) => {
-                  e.preventDefault()
-                  createNewPost(PostStatusType.Draft)
-                }}
-              >
-                Save Draft
-              </Button>
-            </div>
-            <style jsx>{`
+      <DashboardLayout>
```
nlitwin, line 62: Only change in this file is deleting the `{(currentUser) => ()}` wrapper around `<DashboardLayout />`

robin-macpherson:
@nlitwin has each page been run and checked carefully to make sure everything works as expected?

nlitwin:
@robin-macpherson just pushed up a fix for the nav rerendering / reanimating when navigating to the New Post page. It should be good to go! Here is a gif of all the pages logged in plus the post page both logged out and logged in:

logged_in_and_out

robin-macpherson:
@nlitwin looks good! As mentioned in our outside messages, I'll merge this and the next one so that we can proceed with the re-structuring off the app!

PR #163, opened 2020-07-13
Restructure Cleanup (part 1)

Lanny:
## Description

Absorb backend code into frontend directory in a more natural way.

- most code from backend/src has moved into frontend/nexus
- backend's `graphql.ts` file has been split but a bit (more to do there)
- pulled some scripts from root into FE
- got rid of some netlify stuff
- consolidated the three README's into FE's README

What's left is to split up what's left in `nexus/graphql.ts` into files based on graphql type (e.g. thread object def, queries and mutation to `thread.ts`), and moving everything that's under FE to root and consolidating files there (just cypress I think)


robin-macpherson:
@Lanny thanks for splitting up the hefty file!

frontend/pages/api/playground.ts
```diff
@@ -1,9 +1,17 @@
...
+  // Disable playground outside of development
+  handler = (_: any, res: any) => {
+    res.status(403).send('GQL playground is disabled outside of development')
+  }
+}
 
-app.assemble()
-
-export default app.server.handlers.playground
+export default handler
```
robin-macpherson, line 17: Oooh so are we good with that question around configuring the playground page now!?

Lanny: Unfortunately we're not. This is just to disable the playground in prod environments once it works (users don't need it and the polling eats up a lot of requests).

Fixing it is probably a simple change in a prisma repo somewhere but IMO not a top priority since the electron playground app thing exists

Issue #180, opened 2020-07-24
Visual Bugs & Improvements


robin-macpherson:
#### Visual Bugs To Fix

- [x] My Feed search box isn't the right width [#181] 
- [x] Chevron/down arrow is in wrong position on My Feed filters [#181]
- [x] Chevron/down arrow is in wrong position on Settings Add Languages
- [x] Chevron/down arrow is in wrong position on New Post drop downs
- [x] Settings > languages drop downs are off on mobile view

PR #181, opened 2020-07-24
Fix visual bugs in MyFeed search & filters

robin-macpherson:
## Description

**Issue:** #180 

Fixes the visual bugs around My Feed search & filters

## Subtasks

- [x] Fix search box issues
- [x] Add new chevron icons
- [x] Add nice `ChevronDownIcon` to select boxes and style nicely

### Before

image

### After

image


robin-macpherson:
> Approving now, but if possible I'd like to see if it would be possible to have a single chevron icon component that takes a direction prop. Using some like:
> 
> ```
> <g transform={`translate(${rotation}deg)`}>
>   ... rest of the svg
> </g>
> ```
> 
> Should make it possible assuming the directions all look the same save for rotation.

@Lanny good idea! Since timing is so tight and there's a lot to do, for now I've just removed the other `ChevronIcon`s and have just a single `ChevronIcon` when defaults to downward facing for now and can add the rotation functionality in another PR soon

PR #186, opened 2020-07-26
POC: Set up Application State with React Context

### LINKED ISSUE #187: Set up Application State
#### Perceived Problem

- Currently, we do not have a way to manage application-level state and I believe that the currentUser is an excellent use case for having an application state context/provider.

- Create a React Context for managing application level state which is flexible/composable and enables us to simply wrap the custom App in an ApplicationContext and then anywhere in the app that needs currentUser, in this case, can simply call useApplicationState() and destructure the currentUser value.

robin-macpherson:
## Description

**Issue:** closes #187 

Currently, we do not have a way to manage application-level state and I believe that the `currentUser` is an excellent use case for having an application state context/provider.

This POC creates a React Context for managing application level state which is flexible/composable and enables us to simply wrap the custom `App` in an `ApplicationContext` and then anywhere in the app that needs `currentUser`, in this case, can simply call `useApplicationState()` and destructure the `currentUser` value.

I feel that this is a very clean and simple implementation that drastically simplifies things and enables us to simply load the currently logged in user (or `null`) once at the top level and then access that value from anywhere without having to pass it around through many layers of props which feels brittle and error-prone.

This POC uses the `feed` query an its example since that is what I'm working on in a separate PR and require having this `currentUser` value in a way that is currently not well supported and made quite tricky by the way that `<AuthGate>` currently works.

## Subtasks

- [x] Set up the `Context`
- [x] Figure out a nice `Context.Provider` implementation
- [x] Figure out an incredibly simple and easy-to-use custom hook implementation
- [ ] Test and fix any remaining TypeScript / Apollo errors

### DIFF
--- a/frontend/hooks/useApplicationState.ts
+++ b/frontend/hooks/useApplicationState.ts
@@ -1,0 +1,4 @@
+import { useContext } from 'react'
+import { ApplicationContext } from '../state/ApplicationStateProvider'
+
+export const useApplicationState = () => useContext(ApplicationContext)
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -6,6 +6,7 @@ import { initializeTracking, trackPageView } from '../lib/google-analytics'
 import '../styles/reset.css'
 import '../styles/globalStyles.css'
 import { appWithTranslation } from '../config/i18n'
+import { ApplicationStateProvider } from '../state/ApplicationStateProvider'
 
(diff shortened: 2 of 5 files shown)


robin-macpherson:
This is still `WIP` but almost done and posting for visibility.

robin-macpherson:
Closing due to Apollo awesomeness, the branch will come in handy when we eventually use this pattern for things like notifications.

PR #197, opened 2020-07-31
Implement a way to get to the dashboard from landing page when logged in

robin-macpherson:
## Description

Users are very likely to come to Journaly via `journaly.com` but currently it's very confusing as there's no clear path to get to the dashboard unless you know the exact url path to the dashboard. This PR allows for conditionally rendering a dashboard link in the site nav.

image

## Subtasks

- [x] Implement

### DIFF
--- a/frontend/components/Site/Nav/Nav.tsx
+++ b/frontend/components/Site/Nav/Nav.tsx
@@ -1,63 +1,77 @@
 import { width, darkGrey } from '../../../utils'
 import NavLink from '../../NavLink'
 import Logo from '../../Logo'
+import { useCurrentUserQuery, User as UserType } from '../../../generated/graphql'
 
-const navItems = [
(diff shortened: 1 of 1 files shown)


frontend/components/Site/Nav/Nav.tsx
```diff
@@ -1,63 +1,77 @@
...
+  const { data } = useCurrentUserQuery()
+  const currentUser = data?.currentUser as UserType
+
+  return (
+    <div>
+      <div className="header-container">
+        <Logo />
+
+        <ul className="nav-items">
+          <NavLink href="/about" key="About">
```
nlitwin, line 16: We can remove all the `key`s if we aren't looping over with map anymore

robin-macpherson: Good point!

PR #206, opened 2020-08-01
Add update & delete comment functionality and improve comment thread UI

### LINKED ISSUE #113: Expand Commenting CRUD Operations
#### Perceived Problem

- Currently comments only support creation

- Add editing and deleting

**frontend:** #206 
**backend:** #105

robin-macpherson:
## Description

**Issue:** closes #113 

- Builds out the ability to `update` and `delete` actions for comments inside of threads.
- Breaks comments out into their own component
- Drastically improves commenting UI!
  - Adds avatar to comments
  - Makes use of our theme buttons
  - Improves some minor layout/spacing things so that everything is consistent

## Subtasks

- [x] Break `Comment` out into its own component
- [x] bring in and set up update hooks for comments
- [x] bring in and set up delete hooks for comments
- [x] update styling
- [x] test

### The full experience

edit-delete-comment-exp

### DIFF
--- a/frontend/components/InlineFeedbackPopover/Comment.tsx
+++ b/frontend/components/InlineFeedbackPopover/Comment.tsx
@@ -94,7 +94,7 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment }) 
         )}
         {canEdit && isEditMode && (
           <Button size={ButtonSize.Small} onClick={updateExistingComment} disabled={loading}>
-            save
+            Save
           </Button>
         )}
       </div>
--- a/frontend/components/InlineFeedbackPopover/Comment.tsx
+++ b/frontend/components/InlineFeedbackPopover/Comment.tsx
@@ -1,0 +1,146 @@
+import React, { useState } from 'react'
+import {
+  useUpdateCommentMutation,
+  useDeleteCommentMutation,
+  CommentFragmentFragment as CommentType,
+} from '../../generated/graphql'
+import Button, { ButtonSize, ButtonVariant } from '../../elements/Button'
+import BlankAvatarIcon from '../Icons/BlankAvatarIcon'
+import theme from '../../theme'
+
+type CommentProps = {
+  comment: CommentType
+  canEdit: boolean
+  onUpdateComment: any
+}
+
+const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment }) => {
+  const [isEditMode, setIsEditMode] = React.useState<boolean>(false)
+  const [updatingCommentBody, setUpdatingCommentBody] = useState<string>(comment.body)
+
+  const [updateComment, { loading }] = useUpdateCommentMutation({
+    onCompleted: () => {
+      onUpdateComment()
+      setUpdatingCommentBody('')
+    },
+  })
+
+  const updateExistingComment = () => {
+    updateComment({
+      variables: {
+        commentId: comment.id,
+        body: updatingCommentBody,
+      },
+    })
+    setIsEditMode(false)
+  }
+
+  const [deleteComment, { loading: deleteLoading }] = useDeleteCommentMutation({
+    onCompleted: () => {
+      // just refetches the post as in updateComment
+      onUpdateComment()
+    },
+  })
+
+  const deleteExistingComment = () => {
+    deleteComment({
+      variables: {
+        commentId: comment.id,
+      },
+    })
+  }
+
+  return (
+    <div className="comment">
+      <div className="author-block">
+        {comment.author.profileImage ? (
+          <img className="profile-image" src={comment.author.profileImage} alt="" />
+        ) : (
+          <BlankAvatarIcon size={20} />
+        )}
+        <span className="author-identifier">
+          {comment.author.name
+            ? `${comment.author.name} (@${comment.author.handle})`
+            : `@${comment.author.handle}`}
+        </span>
+      </div>
+      <div className="body-block">
+        {isEditMode ? (
+          <textarea
+            value={updatingCommentBody}
+            onChange={(e) => setUpdatingCommentBody(e.target.value)}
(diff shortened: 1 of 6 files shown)


frontend/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+type CommentProps = {
+  comment: CommentType
+  canEdit: boolean
+  onUpdateComment: any
+}
+
+const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment }) => {
+  const [isEditMode, setIsEditMode] = React.useState<boolean>(false)
+  const [updatingCommentBody, setUpdatingCommentBody] = useState<string>(comment.body)
```
nlitwin, line 19: nit: maybe `updatedCommentBody`? It does say `useState<string>`, but I had to double check that it wasn't a boolean

frontend/theme/index.ts
```diff
@@ -58,7 +58,7 @@ const theme: Theme = {
     gray100: '#f6f6f6',
     gray200: '',
     gray300: '',
-    gray400: '',
+    gray400: '#dadada',
```
nlitwin, line 61: 🎶

frontend/components/InlineFeedbackPopover/index.tsx
```diff
@@ -90,7 +90,7 @@ const Popover: React.FC<PopoverProps> = ({ target, children }) => {
           max-height: ${ownHeight}px;
 
           background-color: #f9f9f9;
-          border: 1px solid #dadada;
+          border: 1px solid ${theme.colors.gray400};
```
nlitwin, line 93: 👌

frontend/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -0,0 +1,146 @@
...
+            value={updatingCommentBody}
+            onChange={(e) => setUpdatingCommentBody(e.target.value)}
+          />
+        ) : (
+          <p>{comment.body}</p>
+        )}
+        {canEdit && !isEditMode && (
+          <div className="edit-block">
+            <Button size={ButtonSize.Small} onClick={() => setIsEditMode(true)}>
+              edit
```
nlitwin, line 79: Maybe `Edit` (and `Delete` / `Save` on lines 87 and 93) to match the capitalization of `Submit` and the other buttons?

frontend/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -0,0 +1,146 @@
...
+          margin-left: 5px;
+        }
+
+        .profile-image {
+          width: 27px;
+          height: 27px;
+          object-fit: cover;
+        }
+
+        :global(.author-block svg) {
```
nlitwin, line 115: Since `.author-block` is part of this component, I think we want `.author-block :global(svg) {`

frontend/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -0,0 +1,146 @@
...
+        .body-block {
+          display: flex;
+          justify-content: space-between;
+        }
+
+        .edit-block {
+          display: flex;
+        }
+
+        :global(button:first-child) {
```
nlitwin, line 129: When I did just `:global(X)` in other components, I could have sworn I saw the styles apply to other elements outside of the component I was styling. It might be safer to do `.edit-block :global(button:first-child) {` or give a className to `<Button>`  and just use that class directly (which would also help decrease CSS specificity)

nlitwin:
🔥 👌 🎉  This improves the UX so much! yeeeehawww 🏇

frontend/components/InlineFeedbackPopover/Comment.tsx
robin-macpherson: Hmm only problem is that it's not just a 'past' thing, since this also represents the state when you begin editing, etc. At least we have the nice TS annotation so you can see it's a string! 😅 How about I add a comment above?

frontend/theme/index.ts
robin-macpherson: tee hee 🎷

frontend/components/InlineFeedbackPopover/Comment.tsx
robin-macpherson: Sounds good to meeee, done!

frontend/components/InlineFeedbackPopover/Comment.tsx
robin-macpherson: Great call!

frontend/components/InlineFeedbackPopover/Comment.tsx
robin-macpherson: Sounds good to me! Take a look at the updated 😄

frontend/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -0,0 +1,150 @@
...
+              onClick={deleteExistingComment}
+              disabled={deleteLoading}
+            >
+              Delete
+            </Button>
+          </div>
+        )}
+        {canEdit && isEditMode && (
+          <Button size={ButtonSize.Small} onClick={updateExistingComment} disabled={loading}>
+            save
```
nlitwin, line 97:
```suggestion
            Save
```

frontend/components/InlineFeedbackPopover/Comment.tsx
nlitwin: Oh it's ok, I don't think a comment is necessary 😊  traditionally anything with `-ing` is a boolean, but it's really minor, so that's why i put `nit: ` 🤣 something i've used on other teams for nitpicky suggestions that you can feel free to ignore 😛

PR #209, opened 2020-08-01
🔥 FIRE: Drafts are showing in `My Feed` page and causing server errors and user concern/confusion

robin-macpherson:
## Description

Adds extra condition to `feed` query so that drafts do not show on My Feed page

...

## Subtasks

- [x] Test locally


robin-macpherson:
Merging this in now due to high urgency and I don't think anyone is around to approve for a while

PR #214, opened 2020-08-02
Add Language Badge to PostCards in My Feed 

### LINKED ISSUE #211: POST CARDS: Add more information to `PostCard`s
#### Perceived Problem

- A number of users have noted that it would be super helpful to have more information on the `PostCard`s in the feed:
  - post language

- Add an extra section underneath the `PostCard` for this information

robin-macpherson:
## Description

**Issue:** closes #211 

- A number of users have noted that it would be super helpful to have more information on the `PostCard`s in the feed:
  - post language
- This PR adds a language badge underneath the author information on a `PostCard` only when it is in the feed

**NOTE:** There is also a great use case for displaying the language badge in the other scenarios (`!stacked` and on the profile, for example) but it requires restructuring the component quite a bit and after spending over 30 mins just messing with that, I decided this is WAY higher impact so best to merge this in as is and @nlitwin you could work on the other use cases since you're most familiar with that component. We can discuss the visual design.

## Subtasks

- [x] Update GQL `PostCardFragment`
- [x] Add for `stacked` cards

### Regular

image

### Long language name

image

### DIFF
--- a/frontend/components/Dashboard/PostCard/PostCard.tsx
+++ b/frontend/components/Dashboard/PostCard/PostCard.tsx
@@ -38,6 +38,7 @@ const PostCard: React.FC<Props> = ({
     threads,
     author: { handle, name, profileImage },
     createdAt,
+    language: { name: languageName },
   } = post
   const isDraft = status === PostStatusType.Draft
   const isPublished = status === PostStatusType.Published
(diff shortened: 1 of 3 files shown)


nlitwin:
This might be an edge case to handle later, since we probably don't have mega long language names, but when the screen is larger and more PostCards fit in a row, the language wraps and gets right up to the date/readtime.

robin-macpherson:
@nlitwin good call! Yep I'd say let's handle this later, perhaps in your follow up ticket (did you see where I tagged you in the description?) 

Thanks for the quick approval!

nlitwin:
Yup yup! happy to work on other user cases for the PostCard 😊

Issue #223, opened 2020-08-02
Fix height of sections on the Profile Page


nlitwin:
#### Description

When the height of the screen is short, content goes over the boundary since each section doesn't take up the full height that it needs

![image](https://user-images.githubusercontent.com/5829188/89127597-825b4a00-d4a3-11ea-849c-161b3fec2942.png)

#### Repro Steps

1. Go to the Profile Page
2. Make your screen squat

PR #224, opened 2020-08-02
Allow profile sections to take up more height

nlitwin:
## Description

**Issue:** Closes: #223 

Fixes this described in the issue above: 

## Subtasks

- [x] Make profile details and recent posts take up the height they need to not have their content spill over
- [x] Show `Recent Posts` title on desktop too

### DIFF
--- a/frontend/components/Dashboard/Profile/PostList.tsx
+++ b/frontend/components/Dashboard/Profile/PostList.tsx
@@ -41,7 +41,6 @@ const PostList: React.FC<Props> = ({ currentUserId }) => {
         .post-list {
           display: flex;
           flex-direction: column;
-          justify-content: center;
           align-items: center;
           padding: 0 ${layoutLeftRightPadding} ${layoutTopBottomPadding};
           background-color: ${theme.colors.white};
@@ -57,11 +56,6 @@ const PostList: React.FC<Props> = ({ currentUserId }) => {
         .posts-title {
           margin: 40px 0;
           ${theme.typography.headingLG};
-        }
-        @media (min-width: ${theme.breakpoints.MD}) {
-          .posts-title {
-            display: none;
-          }
         }
 
         .post-list :global(.post-card-container) {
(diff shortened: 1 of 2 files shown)


robin-macpherson:
Hmm let's see, just want to make sure I understand the PR/intent fully. Ideally we want the whole profile to be visible unless the height of the viewport is really low. So visually I would prefer that we constrain the height from expanding beyond a certain point. For example as we chatted about in person yesterday, the problem in the first screenshot ideally shouldn't occur in the first place because the posts should kind of disappear and you should be able to scroll up and down through the post list inside of that right section, does that make sense?

robin-macpherson:
So if I'm understanding the current PR/intent correctly (I definitely might be missing something so feel free to correct me!) I would actually prefer we try to fix the central issue which is that the post list explodes/overflows out of the section and see if we can implement some scrolling that is contained inside the `PostList`

robin-macpherson:
I wonder how quickly we could implement a "Read more..." for bio's that are longer than a certain amount. But the other option that might just be both simpler to implement and also simpler for the user because it sets an expectation is to have a max character length on the bio. We could start by just using whatever the char count is of the demo we have since we've built/designed around that! Whatcha think!? 😄

nlitwin:
@robin-macpherson Thanks for the comments! I'm working on the PostList now to make it scroll if there are more posts than fit the PostList container 👍  We will definitely need to set a max length on the bio though like you said so that the profile details don't get taller than the max-width of the PostList that allows it to scroll.

I think we could add the max length to the bio in another PR, since that max length needs to be enforced in the Settings page when users update their bio there, which isn't wired up yet. What do you think? 😄

nlitwin:
I updated the PR description with a new GIF :D

robin-macpherson:
> @robin-macpherson Thanks for the comments! I'm working on the PostList now to make it scroll if there are more posts than fit the PostList container 👍 We will definitely need to set a max length on the bio though like you said so that the profile details don't get taller than the max-width of the PostList that allows it to scroll.
> 
> I think we could add the max length to the bio in another PR, since that max length needs to be enforced in the Settings page when users update their bio there, which isn't wired up yet. What do you think? 😄

Sounds good to me! I can handle the max length then because I was going to hook up those settings today potentially while adding in the image uploading for profile image! I'll keep you updated on that!

PR #231, opened 2020-08-03
Position thread popover in the right place for new threads.

Lanny:
## Description

Previously starting a new thread would put the thread popover either in the position of the last opened active thread, or at the top left of the screen. It's not positioned relative the the actual section of text that was selected to create a thread.


robin-macpherson:
@Lanny I'm actually seeing some issues in the Vercel preview and I'm not quite sure the cause yet but wanted to bring them up quickly. It looks like this branch is a few days behind master as I'm seeing the pre-pills welcome modal style so I'm not sure if that could be the cause of anything, but I'm also getting server errors:

image

robin-macpherson:
@Lanny just pulled the branch down and merged master into it and will double check the Vercel build adter, and if everything is working great will merge it in

Issue #235, opened 2020-08-04
Create generic Select and MultiSelect components


nlitwin:
#### Perceived Problem

- We have a great LanguageSelect and LanguageMultiSelect, but we'll also need Select/MultiSelect components for things other than languages. We also need to support loading state (see #220) 

#### Ideas / Proposed Solution(s)

- Build generic Select and MultiSelect components that can handle any type of data

Issue #220, opened 2020-08-02
Add disabled state to select component, use it in language select components


Lanny:
#### Perceived Problem

When languages are being fetched, the language select shows up as an empty select, when adding a learning/native language, there's no visual indication that an operation is happening

#### Ideas / Proposed Solution(s)

Adding a disabled state to select/language select components would help make it clear that there's some kind of blocking operation happening and that the user should wait.

PR #236, opened 2020-08-04
Create generic Select and MultiSelect components

nlitwin:
## Description

**Issues:** 
- Closes: #235
- Closes: #220 

We have a great LanguageSelect and LanguageMultiSelect, but we'll need Select/MultiSelect components for things other than languages as well as support for a loading state (see #220). This PR just introduces the generic components themselves, as integrating them into the existing LanguageForms (on the settings page and in the welcome modal) started getting pretty large. The new components are inspired by the existing LanguageSelect components but are tweaked to just be a little more generic.

## Open Question

On the My Feed page, we have selects that have a box-shadow and no border. On the Settings page the selects have borders and no box-shadow. Should we:

1) standardize on one style? ✅ 
2) build in both style versions as options to the component?
3) choose one style as a default and let the parent style it differently, if need be?

My personal preference in order is 1, then 3, then 2 for keeping component variations low and making it easier to maintain components going forward, but I'm open! Watcha think @robin-macpherson ?

## Subtasks

- [x] Create Select
  - [x] Handle `loading` and `disabled` states
- [x] Create MultiSelect
  - [x] Handle `loading` and `disabled` states

## Screenshots

select_multiselect_example


robin-macpherson:
@nlitwin this is a tiny detail but can we have the chevron loading thingy blue? 🤩

nlitwin:
@robin-macpherson awesome! then I think this PR should be ready, since its styles include the box-shadow + no border version 😊

Issue #238, opened 2020-08-05
SETTINGS/PROFILE - User Details Form


robin-macpherson:
#### Perceived Problem

- Users are unable to upload a profile photo
- They also can't update their core information

#### Ideas / Proposed Solution(s)

- Build out profile image uploading and wire up form for updating name, email, display name, ...

PR #239, opened 2020-08-05
SETTINGS/PROFILE: Wire up user details & profile image upload

robin-macpherson:
## Description

**Issue:** #238 

Build out the capability for users to:
- upload a profile image
- update their core information:
  - display nam
  - email
  - name
- I also finally fixed the hover colors for secondary button variant

**NOTE:** Also disables the location field for now since this requires some more work to figure out how we will handle the `Location` table

🚨 **NOTE #2:** 🚨 I'm getting a lot of communications from users about this specific section not working and it seems to be causing a lot of confusion, so I'd like to work to get this PR through ASAP so we can at least have the functionality working. I honestly had quite a lot of trouble trying to mess around with the form and validation stuff and so this definitely needs some clean up and refactoring, and there seems to be a bug with the loading spinner with the secondary button.

**Issues to follow up on:**
- loading spinner has a bug on secondary button variant - looks like a circle
- form validation is not currently working due to trouble getting that working

## Full Flow

profile-update-preview

## Subtasks

- [x] Re-do the form fields to match the DB fields
- [x] Add `updateUser` Mutation (BE)
- [x] Add `updateUser` Mutation (FE)
- [x] Codegen apollo hooks
- [x] Scaffold everything out and get state & handlers wired up
- [x] Get updateUser working with standard inputs
- [x] Get image upload working!
- [x] Test


frontend/components/Dashboard/Settings/DetailsForm.tsx
```diff
@@ -1,13 +1,52 @@
...
+  target: HTMLInputElement & EventTarget
+}
+
+const DetailsForm: React.FC<DetailsFormProps> = ({ currentUser }) => {
+  const [handle, setHandle] = useState(currentUser.handle)
+  const [email, setEmail] = useState(currentUser.email)
+  const [name, setName] = useState(currentUser.name || '')
+  const [profileImage, setProfileImage] = useState(currentUser.profileImage)
+  const [uploadingImage, setUploadingImage] = useState(false)
+  const fileInput = useRef<HTMLInputElement>(null)
```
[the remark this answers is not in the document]
Lanny: oh, maybe some affordance for multiple transforms needs to be there, but it does seem like we should be able to generalize this logic

frontend/components/Dashboard/Settings/DetailsForm.tsx
```diff
@@ -18,21 +57,38 @@ const DetailsForm: React.FC = () => {
...
   const handleDetailsSubmit = (): void => {}
-  const handleSubmitImageClick = (): void => {
-    // TODO save new image
+
+  const handleUpdateUser = (e: React.FormEvent) => {
+    e.preventDefault()
+
+    updateUser({
+      variables: {
+        userId: currentUser.id,
```
Lanny, line 66: Using cookie auth so we shouldn't need to pass this in

frontend/nexus/user.ts
```diff
@@ -85,6 +85,28 @@ schema.extendType({
...
+        profileImage: schema.stringArg({ required: false }),
+      },
+      resolve: async (_parent, args, ctx: any) => {
+        const updates = { ...args }
+        delete updates.userId
+
+        return ctx.db.user.update({
+          data: updates,
+          where: {
+            id: args.userId,
```
Lanny, line 104: Cookie auth again, this will allow anyone to update anyone else's profile

frontend/nexus/user.ts
```diff
@@ -85,6 +85,28 @@ schema.extendType({
...
+        handle: schema.stringArg({ required: false }),
+        email: schema.stringArg({ required: false }),
+        name: schema.stringArg({ required: false }),
+        profileImage: schema.stringArg({ required: false }),
+      },
+      resolve: async (_parent, args, ctx: any) => {
+        const updates = { ...args }
+        delete updates.userId
+
+        return ctx.db.user.update({
```
Lanny, line 101: This can fail since `handle` has the unique constraint. I'm not sure we want to allow users to change their handles, but if we really do then we should handle collisions more gracefully and notify users on the FE if their target handle is taken.

robin-macpherson: Great point! I do want people to be able to change their handle if they would like to at some point but I think it would be best to simply disable this field for now so we can work on a nice experience for checking handle availability

frontend/components/Dashboard/Settings/DetailsForm.tsx
robin-macpherson: Great point!

frontend/nexus/user.ts
robin-macpherson: Done! 😄

frontend/components/Dashboard/Settings/DetailsForm.tsx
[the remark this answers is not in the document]
robin-macpherson: 🙌🏼

frontend/components/Dashboard/Profile/ProfileCard.tsx
```diff
@@ -31,20 +31,34 @@ const ProfileCard: React.FC<Props> = ({ currentUser }) => {
 
   const name = currentUser.name || currentUser.handle
   const showSeparator = user.facebook || user.instagram || user.youtube || user.website
+  const profileImage = currentUser.profileImage
+  const speaks = []
```
nlitwin, line 35: if you want to do it in one line you could do `const speaks = currentUser.langaugesNative.map(({ language }) => language.name)` 😄

robin-macpherson: Oh right! It's funny, I originally started writing it as a Python list comprehension and then was like awww can't do those, and forgot about good ol' `map` haha 😂

robin-macpherson: Eh, everything is blowing up with TS when I do that and I'm exhausted 😅🙈

PR #240, opened 2020-08-05
Add `publishedAt` field to Posts

### LINKED ISSUE #213: DATABASE: We should store `publishedAt` in addition to `createdAt`
#### Perceived Problem

- Currently we just rely on the `createdAt` field but this is inaccurate because a post could sit as an `edit` for any length of time

- Store `publishedAt` upon publishing a post

robin-macpherson:
## Description

**Issue:** closes #213 

Adds `publishedAt` field to `Post`s so that we can correctly record when a post was published, not simply created.

**Production data migration plan:** copy values in `createdAt` to `publishedAt` for all existing posts that have a `status` of `'PUBLISHED'`.

## Subtasks

- [x] Update DB schema
- [x] Migrations
- [x] Update resolvers
- [x] Use new field in client
- [x] Prepare to migrate production data


frontend/components/Dashboard/PostCard/PostCard.tsx
```diff
@@ -89,7 +90,8 @@ const PostCard: React.FC<Props> = ({
                 )}
 
                 <div className="post-subtext">
-                  {formatShortDate(createdAt)} - {t('readTime', { minutes: readTime })}
+                  {publishedAt ? formatShortDate(publishedAt) : formatShortDate(createdAt)} -{' '}
```
Lanny, line 93:
```suggestion
                  {formatShortDate(publishedAt || createdAt) -{' '}
```

frontend/nexus/post.ts
```diff
@@ -144,7 +145,7 @@ schema.extendType({
           skip: args.skip,
           first: args.first,
           orderBy: {
-            createdAt: 'desc',
+            publishedAt: 'desc',
```
Lanny, line 148: how does this handle `null` in the publishedAt col? Do they end up at the start or end of the list?

frontend/nexus/post.ts
```diff
@@ -264,6 +267,10 @@ schema.extendType({
           data = { ...data, ...processEditorDocument(args.body) }
         }
 
+        if (args.status === 'PUBLISHED' && !originalPost.publishedAt) {
+          data = { ...data, publishedAt: new Date().toISOString() }
```
Lanny, line 271:
```suggestion
          data.publishedAt = new Date().toISOString()
```

Doesn't affect anything other than not having to copy the whole `data` object

frontend/nexus/post.ts
robin-macpherson: They end up at the beginning, that's why I'm proposing that we populate the `publishedAt` column with the same value as `createdAt` for all existing posts that have a `status` of `'PUBLISHED'` so that we can preserve the same ordering as there is today. Does that sound good to you?

frontend/components/Dashboard/PostCard/PostCard.tsx
Lanny: sorry, missing curly brace after closing paren

frontend/nexus/post.ts
Lanny: yup, that works for me 👍

frontend/components/Dashboard/PostCard/PostCard.tsx
robin-macpherson: Got it ✅

PR #245, opened 2020-08-06
Update details form validations

nlitwin:
## Description

Quick fix for form validation errors in the details form of the settings page. This PR is pointed into #239 so that it's easier to see the diff and not just commit directly to the PR.

## Screenshots

image

### DIFF
--- a/frontend/components/Dashboard/Settings/DetailsForm.tsx
+++ b/frontend/components/Dashboard/Settings/DetailsForm.tsx
@@ -18,6 +18,7 @@ interface HTMLInputEvent extends React.FormEvent {
 }
 
 const DetailsForm: React.FC<DetailsFormProps> = ({ currentUser }) => {
+  const [handle, setHandle] = useState(currentUser.handle)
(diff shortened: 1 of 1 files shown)


frontend/components/Dashboard/Settings/DetailsForm.tsx
```diff
@@ -112,29 +109,27 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ currentUser }) => {
...
               <div className="details-form-field">
                 <label className="settings-label" htmlFor="handle">
-                  Display Name
+                  {t('profile.details.displayNameLabel')}
                 </label>
                 <input
                   type="text"
                   name="handle"
-                  value={handle}
+                  value={currentUser.handle}
```
nlitwin, line 117: Since this field is disabled, but we still have it wired up to be able to change the value with `setHandle`, I removed the `useState` for handle in case someone opens the dev tools and removes the `disabled` attribute from the input element

PR #256, opened 2020-08-07
Stop logging current user on Feed

robin-macpherson:
## Description

We should stop console logging the current user in the browser on the my feed page. It will be simple to bring `currentUser` back when we start using it.

## Subtasks

- [x] Clean up

### DIFF
--- a/frontend/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/frontend/components/Dashboard/MyFeed/MyFeed.tsx
@@ -32,9 +32,6 @@ const MyFeed: React.FC<Props> = () => {
   const showPagination = count > NUM_POSTS_PER_PAGE
   const pageTitle = 'My Feed'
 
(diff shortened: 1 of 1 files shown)


frontend/components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -32,8 +32,9 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
   const showPagination = count > NUM_POSTS_PER_PAGE
   const pageTitle = 'My Feed'
 
+  console.log(count)
```
nlitwin, line 35: I should have the pagination bug fixed shortly so we can remove this console log and potentially the comment on line 37, if you want!

robin-macpherson: Cool cool! Done!

Issue #255, opened 2020-08-07
🐞 Pagination calculation seems to be off


robin-macpherson:
#### Description

![image](https://user-images.githubusercontent.com/34203886/89609224-11fa5300-d82c-11ea-93a7-6a6da4f7d8a9.png)

The pagination calculation seems to be off. We tend to get a the next page around 1 post too early and then end up with an empty last page.

Issue #208, opened 2020-08-01
Read time calculation improvements in ideographic writing systems


Lanny:
#### Perceived Problem

Currently we estimate the number of words in a post as the number of spaces plus one, and assume a reading speed of 200 words per minute. This fails pretty badly in languages that don't use spaces as word separators. 

#### Ideas / Proposed Solution(s)

- While there are techniques for best-guess word separation in languages with idiographic writing systems, we can probably just store info in the languages table that indicate if a language has a idiographic writing system and find a rough characters-per-minute estimate for those languages.

PR #261, opened 2020-08-07
Fix number of pages for a given number of posts

nlitwin:
## Description

**Issue:** Closes: #255 

This PR aligns the count query with the posts query on the My Feed page (drafts were omitted from the posts query, so the count query needs it too)

## Subtasks

- [x] Update `PostCard` to have min width
- [x] Update `readTime` to not show 0 minutes (show minimum of 1 min.)

### Min-width on the PostCard makes this not happen 😄 

image

### DIFF
--- a/frontend/nexus/utils.ts
+++ b/frontend/nexus/utils.ts
@@ -53,9 +53,11 @@ const getNodeTagName = (node: NodeType): string => {
 }
 
 const isEmptyParagraph = (node: NodeType): boolean => {
-  if (node.type !== 'paragraph') return false
+  if (node.type !== 'paragraph')
+    return false
 
-  if (node.children && node.children.length > 0) return false
+  if (node.children && node.children.length > 0) 
+    return false
 
   return true
 }
@@ -98,7 +100,9 @@ const extractBodyTextFromNode = (node: NodeType) => {
     return ''
   }
 
-  const content: string = (node.children || []).map(extractBodyTextFromNode).join(' ')
+  const content: string = (node.children || [])
+    .map(extractBodyTextFromNode)
+    .join(' ')
 
   return content
 }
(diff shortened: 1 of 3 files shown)


frontend/nexus/utils.ts
```diff
@@ -145,7 +137,7 @@ export const generateExcerpt = (
 export const readTime = (text: string): number => {
   const numWords = text.split(' ').length
 
-  return Math.round(numWords / 200)
+  return Math.round(numWords / 200) || 1
```
nlitwin, line 140: Changes in file are from prettier except this line

nlitwin: @robin-macpherson Since this gets saved to the DB, it means that future new posts or edited posts will get the "minimum 1 min. read time", but previous posts by alpha testers will still see 0 min. Should I also add an `|| 1` on the frontend for the few alpha tester posts, or just keep it like this for simplicity?

robin-macpherson: > Changes in file are from prettier except this line

I've honestly been finding the prettier stuff pretty annoying to be honest 😕

robin-macpherson: > @robin-macpherson Since this gets saved to the DB, it means that future new posts or edited posts will get the "minimum 1 min. read time", but previous posts by alpha testers will still see 0 min. Should I also add an `|| 1` on the frontend for the few alpha tester posts, or just keep it like this for simplicity?

I was actually assuming that for the moment this would just be a simple frontend change for now since the primary urgency is just presentational and we may want to think more carefully about how we want to represent this in the DB

robin-macpherson: (for example, we may still want the real result of the calculation represented in the data that we store. We may want to run analysis on how many posts are created with 0 minutes, and this also currently helps us identify posts that are impacted by #208

nlitwin: > > Changes in file are from prettier except this line
> 
> I've honestly been finding the prettier stuff pretty annoying to be honest 😕

Really?? I don't think I could live without it now, but I think we should add [some infra via husky](https://prettier.io/docs/en/precommit.html) or something that way files don't get saved _without_ prettier, causing larger file changes when someone edits that file

nlitwin: > > @robin-macpherson Since this gets saved to the DB, it means that future new posts or edited posts will get the "minimum 1 min. read time", but previous posts by alpha testers will still see 0 min. Should I also add an `|| 1` on the frontend for the few alpha tester posts, or just keep it like this for simplicity?
> 
> I was actually assuming that for the moment this would just be a simple frontend change for now since the primary urgency is just presentational and we may want to think more carefully about how we want to represent this in the DB

I agree! will update

Issue #265, opened 2020-08-07
Support RTL languages in Posts


nlitwin:
#### Perceived Problem

Right now, RTL languages are left-aligned. At the very least, they should be right-aligned in Posts. Updating the editor to handle writing from right to left could be tricky, but we should explore how hard/easy it would be to support that too

#### Ideas / Proposed Solution(s)

Support RTL languages as much as possible!

PR #266, opened 2020-08-08
Use dir="auto" to format RTL languages

nlitwin:
## Description

**Issue:** #265 

Display right-to-left languages correctly on PostCards and the Post page

## Subtasks

- [x] Display rtl!

## Screenshots

image
image


Lanny:
Good stuff! Saw that arabic post in prod and got worried, glad to see it ended up looking great without needing to dig into slate internals!

PR #273, opened 2020-08-11
PROFILES: Allow Users to view other users' profiles

### LINKED ISSUE #188: View another user's profile details and posts
#### Perceived Problem

Right now we have profiles at `/dashboard/profile` where the logged in user can view their own profile page. However, there is no way to view _another user's profile page_.

We can add an id to the URL, e.g. `/dashboard/profile/42`, so that we can query that user's profile and posts. If the `currentUser`'s id matches the one in the URL, we can simply display the information already available on `currentUser`, otherwise, we'll query for the profile and posts of the user id in the URL and display that information.

robin-macpherson:
## Description

**Issue:** closes #188 

Currently users can only view their own profiles. This PR refactors things to create a profile api route that uses a new `userById` query to fetch the correct profile.

**FEATURE NOTE:** I decided not to wrap this page in `AuthGate` as it could be nice for anyone to be able to view these nice, fancy profiles, and the posts are also already public so it all makes sense and could make people want to sign up if they get here!

**ROLLOUT NOTE:** This implements all of the functionality so that everything works exactly the same today. I tested that it also works for visiting other profiles and it works perfectly! There will be an immediate follow-up ticket to add a feature that acts as the first entry point to viewing another user's profile 😊

## Subtasks

- [x] Write `userById` query
- [x] Create new API route
- [x] Write new GQL files and generate Apollo hook
- [x] Refactor components to use `user`
- [x] Get all that working
- [x] Add links to visit another user's profile


robin-macpherson:
Tested it out with the preview link and seems to be working splendidly! 😄

frontend/components/Dashboard/Profile/Profile.tsx
```diff
@@ -6,14 +6,14 @@ import { User as UserType } from '../../../generated/graphql'
 import theme from '../../../theme'
 
 type Props = {
-  currentUser: UserType
+  user: UserType | any
```
nlitwin, line 9: Is the `any` because `userById` has extra language stuff on it that `UserType` doesn't have?

robin-macpherson: Still not 100% sure, I tried a whole bunch of things but was having a lot of trouble with it and I saw that we had to do the same thing with Posts

robin-macpherson: I tried also using `UserWithLanguagesFragmentFragment as UserType` for example but yeah nothing that I tried worked

robin-macpherson: Going to merge for now since you approved as I'm working on some stuff that needs this, but if you end up having any more thoughts on it we can update this in my follow-up! 🙂

PR #280, opened 2020-08-12
Order "Recent Posts" on profile page by publish date.

Lanny:
All in the title

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,7 @@
 DATABASE_URL='postgresql://<user>:<password>@localhost:5432/<db_name>'
 APP_SECRET='journaly'
 SITE_DOMAIN='localhost:3000'
+DEPLOYMENT_URL='localhost:3000'
 MAIL_HOST='<mail-host>'
 MAIL_PORT='<port>'
 MAIL_USER='<mail-user>'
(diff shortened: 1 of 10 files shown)


Lanny:
While I was here, I pulled the user and posts queries together into a new "profile" query, so we only have to make one query to load profile pages. In doing that I noticed that we were making these queries on the FE, even on a reload, which shouldn't be happening with SSR. Looks like when we relativized the graphql url it actually broke SSR (my bad, there was even a comment saying not to do that), so this includes logic to fix that. It introduces new env vars to do that, but vercel will orchestrate setting them, and we guess `localhost:3000` as the host in local so it shouldn't require any change to env files.

robin-macpherson:
> While I was here, I pulled the user and posts queries together into a new "profile" query, so we only have to make one query to load profile pages. In doing that I noticed that we were making these queries on the FE, even on a reload, which shouldn't be happening with SSR. Looks like when we relativized the graphql url it actually broke SSR (my bad, there was even a comment saying not to do that), so this includes logic to fix that. It introduces new env vars to do that, but vercel will orchestrate setting them, and we guess `localhost:3000` as the host in local so it shouldn't require any change to env files.

Got it thanks @Lanny ! Wow fascinating. Nice job finding this!

robin-macpherson:
@Lanny while testing the preview link, I noticed I'm getting the unauthenticated dashboard Nav while viewing a post:
image

Lanny:
yup, I'm seeing the same thing. Looking into it now

robin-macpherson:
> yup, I'm seeing the same thing. Looking into it now

Weird, it seems like I'm actually losing my token when clicking into a post so seeing the unauthenticated view because I am indeed logged out

PR #281, opened 2020-08-13
THREADS: Proposal for deleting empty threads

### LINKED ISSUE #210: COMMENT THREADS: Thread should disappear if there are no comments inside of it
#### Perceived Problem

- Currently, if someone decides they don't want to comment on that place or accidentally chose the wrong selection, etc. the comment thread will still remain and there's no way to delete it.

- Add a "cancel" button to the commenting popover component
- Remove thread when popover is closed and `comments.length === 0`

robin-macpherson:
## Description

**Issue:** closes #210 

This is a proposal for a simple (perhaps temporary) solution to removing empty threads. Rather than deferring the DB `save` operation of a thread until after a comment is created, we could simply have a `deleteThread` mutation and when a user closes a thread, we quickly check the length of comments and if it is 0, delete the thread.

## Subtasks

- [ ] ...

### DIFF
--- a/nexus/graphql.ts
+++ b/nexus/graphql.ts
@@ -77,6 +77,34 @@ schema.mutationType({
         })
       },
     })
+    t.field('deleteThread', {
+      type: 'Thread',
+      args: {
+        threadId: intArg({ required: true }),
+      },
+      resolve: async (_parent, args, ctx) => {
+        const thread = await ctx.db.thread.findOne({
+          where: {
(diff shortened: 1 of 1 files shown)


robin-macpherson:
@Lanny just adding the mutation for now to get thoughts and can quickly build out the rest or adjust things if there's any path here for a solution

robin-macpherson:
Proposal approved so moving this over to my other PR ✅

Issue #267, opened 2020-08-09
Float the Journaly editor toolbar as the user scrolls down their new post if it's long


nlitwin:
#### Perceived Problem

If a user is writing a long post, they might want to have access to the journaly editor toolbar without having to scroll up the screen.

#### Ideas / Proposed Solution(s)

Let's investigate what it would entail to have a floating toolbar once the post is long enough and the user has scrolled down

PR #284, opened 2020-08-13
Float the editor toolbar when scrolled down the new post page

nlitwin:
## Description

**Issue:** Closes: #267 

If the user is writing a long post and is scrolled down the page, float the editor toolbar so that it's available to use no matter whether the user is scrolled down on the page

## Subtasks

- [x] Fix the editor toolbar when the user scrolls past it
- [x] Center it on the screen, but pushed over the size of the nav (mobile, mid, and large desktop)

### DIFF
--- a/components/Dashboard/Header/Header.tsx
+++ b/components/Dashboard/Header/Header.tsx
@@ -24,7 +24,7 @@ const Header: React.FC<Props> = ({ onMenuClick }) => {
           display: flex;
           justify-content: space-between;
           align-items: center;
-          height: ${headerHeight}px;
+          height: ${headerHeight};
           padding: 20px 0;
           color: white;
           background-color: ${theme.colors.charcoal};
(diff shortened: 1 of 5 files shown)


nlitwin:
@robin-macpherson PR is ready to look at! I tried over and over to capture GIFs to show tablet/desktop, but every time the file comes out over 10MB. Would you mind playing around with it on the preview link? 😁

nlitwin:
Managed to get a desktop version of the GIF that's really small 🤣

nlitwin:
The floated toolbar should stay centered even if you change the screen width from large to medium to small and back (just couldn't capture that in the GIF without going over 10MB)

nlitwin:
just pushed up a quick small commit that makes the toolbar "pop out" on tablet/desktop more smoothly, so it looks even better now! 🕺

PR #286, opened 2020-08-13
Prevent PostCards from shrinking

nlitwin:
## Description

Safari was shrinking flex items where other browsers were not. Adding `flex-shrink: 0` prevents this from happening

## Before

image

## After

image


nlitwin:
Closing PR since @robin-macpherson said it looked fine

nlitwin:
JK

PR #287, opened 2020-08-15
Use generic Select components in new Language form components

nlitwin:
## Description

**Issue:** Closes: #235 

This PR builds off of the new generic `Select` and `MultiSelect` components and the Language select components that @Lanny created. `NativeLanguageFormField` and `LearningLanguageFormField` components are wrappers around the generic `<MultiSelect />` component that handle interacting with the API to add/remove languages, handle loading states, etc. This allows the Welcome Modal to share these components without duplicating the API and form logic from the languages form in the settings page. Both the Welcome Modal and languages form on the settings page now load the selected languages before showing them so that the prefilled values don't appear a second after the form is already rendered.

## Subtasks

- [x] Create self-contained `Native` and `LearningLanguageFormField` components
  - [x] Handle loading/disabled states
  - [x] Refresh cache after updating (so you can add a language in the welcome modal, go to settings, and see the languages you just updated)
- [x] My Feed page
  - [x] Swap out native `<select>` element for new `Select` component (no new functionality)
  - [x] Remove render prop wrapper since `currentUser` is longer being used in `MyFeed.tsx`
- [x] GraphQL
  - [x] Create `graphql/settingsFormData.graphql` to help with loading settings form data before displaying it
- [x] Settings page
  - [x] Update Languages form to use new native/learning language form field components
- [x] New Post page
  - [x] Use new generic `<Select />` for the new post's language
  - [x] Add missing translation namespace
- [x] Post page
  - [x] Display dialect with language name to fix duplicate key React error message
- [x] Create new language util file and helper function to format languages with dialects `name (dialect)`
- [x] `<Select />`
  - [x] Add docs
- [x] `<MultiSelect />`
  - [x] Update component to use values/ids for the selected options
  - [x] Add docs
- [x] Welcome modal
  - [x] Preload the language data and show the modal once the language data is available
  - [x] Don't show the modal if there is an error fetching languages
  - [x] Replace temporary copy/paste language selects with new native/learning language multiselect components

### Settings Page

native_learning_languages_settings_form

### My Feed using new components (no functionality change)

image

### Welcome Modal

image

### Empty Select on the New Post page

image

### Filled Select on the New Post page

image

### Updated PostAuthorCard

image


pages/dashboard/my-feed.tsx
```diff
@@ -24,12 +24,10 @@ const MyFeedPage: NextPage<InitialProps> = () => {
 
   return (
     <AuthGate>
-      {(currentUser) => (
```
robin-macpherson, line 27: @nlitwin Sorry I'm sure I'm just missing this somewhere but browsing up and down a few times I can't seem to find it, where did `currentUser` move to?

robin-macpherson: It looks like it's just been removed from this level of my feed, but we're going to need it for various things such as their stats bar soon just as an example.

nlitwin: It used to just be a `console.log(currentUser)` in `MyFeed.tsx`, which was taken out recently, so I just removed it here too. I tend to remove unused code, cause it feels weird to leave around and it's easy to add back, but I also don't have a strong preference, I can put it back if you'd like! 😄

robin-macpherson: I generally agree but we're moving super fast right now and so there are a few things here and there that aren't perfect but I don't think particularly bad, and in this case it was left for a reason as we'd discussed previously when I removed the `console.log()`. Please put it back as I'll be using it very soon

robin-macpherson: Also I've mentioned this a few times before but when there are random things that just get removed or changed that aren't related to the PR, I think it's really helpful to just mention them with a comment because in this case this jumped out at me and then I was reviewing the PR with the assumption that this was a part of the PR and in the back of my mind was looking for where this would connect. If I knew right away this was not related I wouldn't have been looking for that connective piece while reviewing the rest

nlitwin: Cool i'll put it back! 😊 I made sure to mention it in the description under `My Feed`: `Remove render prop wrapper since currentUser is longer being used in MyFeed.tsx` that way the context was there. I can move that type of context to an actual code comment instead of the PR description in the future, so that it's easier to see 👍

PR #289, opened 2020-08-15
DESIGN UPDATES 👓 📐♥️👌🏼

robin-macpherson:
## Description

We've built lots of new stuff which is AMAZING! Along the way, it's been necessary to make quick decisions here and there or just get something that works and looks ok. This morning, I took the time to do a wee design audit to compare the overall look & feel to my original mood board and see if I could bring us back inline with that.

This PR is the result of that! It is a collection of small and simple improvements to rather fine details, but I believe it really improves things so much! 🙌🏼

I also took the opportunity to finally implement a solution for cancelling an empty thread (MUCH requested by our users!) 🙌🏼 🙌🏼 🙌🏼

## Subtasks

- [x] Add dark button variant
- [x] Update post card proportions
- [x] Update the comment icon
- [x] Add iconography to comments
- [x] Revamp post author card
- [x] Add cancel button to comments
- [x] ADD CANCEL EMPTY THREADS

#### 1. New `PostCard` proportions so you can see more of the image now
new-design-1

#### 2. Increased font-size in posts with lighter font-weight
new-design-2

#### 3. Introduced iconography to comment editing, new dark button variant, improved author card
new-design-3

#### 4. CANCEL THREAD when it is empty
new-design-4

#### 5. New iconography interactions
new-icons-preview

#### 6. New, simpler and more minimalistic comment icon
new-design-5


components/Dashboard/Post/PostComment.tsx
```diff
@@ -163,8 +163,8 @@ const PostComment: React.FC<PostCommentProps> = ({
 
         .body-block {
           display: flex;
-          justify-content: space-between;
-          flex: 1;
+          overflow-wrap: anywhere;
```
nlitwin, line 166: [This CSS property](https://caniuse.com/#feat=mdn-css_properties_overflow-wrap_anywhere) isn't supported in Safari or IE, would `word-wrap: break-word;` work? (say that 5 times fast 😆 )

components/Dashboard/Post/PostComment.tsx
```diff
@@ -59,57 +61,83 @@ const PostComment: React.FC<PostCommentProps> = ({
...
+            <DeleteIcon size={24} />
+          </span>
+        </div>
+      )}
+      {canEdit && isEditMode && (
+        <>
+          <Button
+            size={ButtonSize.Small}
+            onClick={updateExistingComment}
+            disabled={loading}
```
nlitwin, line 107: I think we want `loading={loading}` here and on line 121, so that the user can see the loading icon while it updates. In `Button.tsx`, it uses `disabled={loading || disabled}`, so the button is already disabled when `loading` is truthy

components/Dashboard/Post/PostComments.tsx
```diff
@@ -71,7 +71,12 @@ const PostComments: React.FC<PostCommentsProps> = ({
               onChange={(e) => setPostCommentBody(e.target.value)}
               disabled={loading}
             />
-            <Button type="submit" disabled={loading} className="submit-btn">
+            <Button
+              type="submit"
+              disabled={loading}
```
nlitwin, line 76: Same as previous comment
```suggestion
              loading={loading}
```

components/Dashboard/Post/PostComments.tsx
```diff
@@ -71,7 +71,12 @@ const PostComments: React.FC<PostCommentsProps> = ({
               onChange={(e) => setPostCommentBody(e.target.value)}
               disabled={loading}
             />
-            <Button type="submit" disabled={loading} className="submit-btn">
+            <Button
+              type="submit"
```
nlitwin, line 75: After submitting, if we call `refetch` then I think the comment should appear after submitting (right now it looks like you have to refresh the page). We'd just need to pass down `refetch` from `pages/post/[id]/index.tsx` to `<PostComments />`. In case you want to do that, you can see an example in `components/Modals/WelcomeModal/WelcomeModalBody.tsx` for how to make TypeScript happy for the `refetch` prop

robin-macpherson: This is actually already happening 😄 Those `onNewComment`, `onUpdateComment`, and `onDeleteComment` props are actually all just `refetch` calls being passed down 🙂 There's no refresh necessary to see new comments

components/Dashboard/Post/PostComment.tsx
robin-macpherson: Great call! Just added that! 🙌🏼 ✅

components/Dashboard/Post/PostComments.tsx
robin-macpherson: 🙌🏼 ✅

components/Dashboard/Post/PostComment.tsx
robin-macpherson: No I tried that and unfortunately it doesn't work 🙁 Good find re: `overflow-wrap` not being supported though!

components/Dashboard/Post/PostAuthorCard.tsx
```diff
@@ -14,35 +14,54 @@ type PostAuthorCardProps = {
...
+        <ul className="language-list">
+          {languagesLearning.map(({ language }) => {
             return <li key={language.id}>{languageNameWithDialect(language)}</li>
           })}
         </ul>
       </div>
+      <div className="stats">
+        <p className="author-info-heading">Has written</p>
+        <ul>
+          <li>10 posts</li>
```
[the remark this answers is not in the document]
robin-macpherson: We might do! I may just leave it for now but good call I'll definitely revisit this card a lot soon. Just wanted to get it looking a wee bit sharper for now! 😄

components/Dashboard/Post/PostComments.tsx
nlitwin: awesome!! even better 😄 i couldn't get it to work locally, but I think that was something weird with my local state

robin-macpherson:
@nlitwin thanks for the reviewwwww!!!

Regarding your ideas:
1. Yeah I'd be up for that! 
2. Great point! I honestly expected some push back on the name "dark" lol I wasn't sure if that would be a viable name with the conventions but couldn't think of anything else in the moment. I just updated everything to `PrimaryDark`! 😄

components/Dashboard/Post/PostComments.tsx
robin-macpherson: Weehoo! 😄 🙌🏼

nlitwin:
overflow_update

Issue #172, opened 2020-07-17
Settings page sections should actually update the user's info


robin-macpherson:
#### Perceived Problem

- Currently, only the languages section is actually wired up to update the user

#### Ideas / Proposed Solution(s)

- Let's make them all work!

Closes #238

PR #290, opened 2020-08-16
View and update bio

nlitwin:
## Description

**Issue:** Part of #172 

This PR allows a user to view and update their bio 🙌🏼 It also adds minor things like displaying the dialect of languages on the `ProfileCard` and a toast success message after updating the bio. When submitting the bio form, it updates so fast that it's easy to not see the loading state of the button, so the toast message should help with that. If that pattern looks good, we can add toast success messages for the rest of the settings page updates, since it's a similar situation of updating fast and not being 100% sure if it worked or not.

## Subtasks

- [x] Prefill bio in the bio form on the settings page and wire up the form for updating it
- [x] Show bio on the `ProfileCard`
- [x] Display dialect with languages on the `ProfileCard`
- [x] Show success and error toast messages after updating bio 

### Updating your bio

updating_bio

### DIFF
--- a/components/Dashboard/Settings/BioForm.tsx
+++ b/components/Dashboard/Settings/BioForm.tsx
@@ -16,6 +16,8 @@ type FormValues = {
   bio: string
 }
 
+const BIO_MAX_LENGTH = 400
+
 const BioForm: React.FC<Props> = ({ bio }) => {
   const { t } = useTranslation('settings')
   const [updateUser, { loading }] = useUpdateUserMutation({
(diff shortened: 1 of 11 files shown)


components/Dashboard/Settings/BioForm.tsx
```diff
@@ -23,13 +50,21 @@ const BioForm: React.FC = () => {
...
             </label>
             {/* TODO: add native maxlength attribute when we know how long this field can be */}
-            <textarea rows={4} id="bio" name="bio" className="j-textarea" ref={register()} />
+            <textarea
+              rows={4}
+              id="bio"
+              name="bio"
+              className="j-textarea"
+              defaultValue={sanitize(bio)}
+              ref={register()}
```
robin-macpherson, line 59:
```suggestion
              ref={register()}
              maxlength="400"
```

PR #319, opened 2020-08-26
Ensure 2 columns of same size

### LINKED ISSUE #310: PostCards on the My Posts page don't stay the same size
#### Description

image

#### Repro Steps

1. Go to the My Posts page
2. Resize the screen

nlitwin:
## Description

**Issue:** Closes #310 

This PR also fixed `PostCard`s that appear very narrow on large width screens. See before and after screenshots below

## Subtasks

- [x] Fix unequal width columns
- [x] Fix narrow `PostCard`s that don't have a long enough title or excerpt
- [x] Fix image height for non-stacked cards

### DIFF
--- a/components/Dashboard/MyPosts/MyPosts.tsx
+++ b/components/Dashboard/MyPosts/MyPosts.tsx
@@ -75,6 +75,7 @@ const MyPosts: React.FC<Props> = ({ currentUser, status }) => {
 
         @media (min-width: ${theme.breakpoints.SM}) {
           .my-posts {
+            /* Ensure both columns have equal width */
             grid-template-columns: repeat(2, minmax(0, 1fr));
           }
         }
--- a/components/Dashboard/MyPosts/MyPosts.tsx
+++ b/components/Dashboard/MyPosts/MyPosts.tsx
@@ -85,7 +85,7 @@ const MyPosts: React.FC<Props> = ({ currentUser, status }) => {
           }
 
           .my-posts :global(.post-card-container) {
-            max-width: 768px;
+            width: 768px;
             margin: 0 auto;
           }
         }
--- a/components/Dashboard/MyPosts/MyPosts.tsx
+++ b/components/Dashboard/MyPosts/MyPosts.tsx
@@ -75,7 +75,7 @@ const MyPosts: React.FC<Props> = ({ currentUser, status }) => {
 
         @media (min-width: ${theme.breakpoints.SM}) {
           .my-posts {
-            grid-template-columns: 1fr 1fr;
+            grid-template-columns: repeat(2, minmax(0, 1fr));
           }
         }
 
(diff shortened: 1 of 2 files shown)


components/Dashboard/MyPosts/MyPosts.tsx
```diff
@@ -75,7 +75,7 @@ const MyPosts: React.FC<Props> = ({ currentUser, status }) => {
 
         @media (min-width: ${theme.breakpoints.SM}) {
           .my-posts {
-            grid-template-columns: 1fr 1fr;
+            grid-template-columns: repeat(2, minmax(0, 1fr));
```
nlitwin, line 78: Apparently 1fr is [not guaranteed to be equal width](https://stackoverflow.com/a/61240964/2918915)!

robin-macpherson:
@nlitwin awesome! Would it be possible to also tack onto this PR fixing the "drafts" tab up a bit? I noticed today in prod (you can see it in the YT video screencast) that the post cards in draft are super narrow and 1 per line it looks quite awkward. If it's a super quick and easy update maybe this PR would be a good place!

nlitwin:
@robin-macpherson for sure! just updated 😊

robin-macpherson:
> @robin-macpherson for sure! just pushed up changes and updated the PR description 😊

Nice thank you! It still looks very awkward to me, how the image is super narrow and out of proportion. I'm actually quite confused by what happened here with the dev of this page. It looks like the drafts are mostly as designed in that they occur one per row, but the proportions are just off that's all. However, I don't know how the published posts ended up with a different implementation where they have multiple per row and look more like post cards from the feed?

robin-macpherson:
@nlitwin for context, here's the original wireframe we were going off of:

image

robin-macpherson:
@nlitwin it's similar to how the post cards on the profile also ended up really off from the designs with regards to proportions in a number of ways:

image

robin-macpherson:
@nlitwin and then here's how those look in prod, even with the published posts:

image

image

nlitwin:
@robin-macpherson I just took a look, and I believe nothing has changed with the `PostCard` layout except [this change](https://github.com/Journaly/journaly/pull/289/files#diff-6d65d32ed10719ea311b7a6b6008d05d) to the image height in #289 (The change for showing more of the image in the PostCards for the My Feed page). I just pushed a commit, which adds back the original image height for the cards when they are in the horizontal layout (not stacked)

robin-macpherson:
@nlitwin I was referring to the original dev on these things. The commit you're referring to there was me trying to fix the design discrepancy that was originally there in the stacked `PostCard`s, but the proportions have been off with the designs this whole time and I think that's what has caused quite a lot of these issues with things not behaving/looking as expected

robin-macpherson:
So for example I was trying to illustrate that I'm having trouble understanding why there is different behavior/layout between "published" and "drafts", although neither one is quite like the designs. Drafts is closer in that it's 1 post per row but the proportions are completely off; and they're also completely off in the profile which is of course the same problem just showing in both places.

Ideally, both tabs should behave/look the same way (1 per row) and the proportions should be as in the designs

robin-macpherson:
@nlitwin as a reference for that, here is how the profile cards were designed with the proportions:
image

And here's how they were built, which the commit you referenced was trying to remedy:
image

Issue #247, opened 2020-08-06
TOPICS: Build Topic Select


robin-macpherson:
#### Ideas / Proposed Solution(s)

Build upon our generic select/multiselect components to create a "popover" that supports:
- typing
- should be able to talk to the server
- server performs fuzzy matching and returns a subset of matching topics

### Tasks
- Previous art?
  - https://react-select.com/home

## REQUIRED

- make a decision on how to handle translations with regards to topics & languages

robin-macpherson:
Closing this one as I think we've gone in another direction over time with a much better topic list and continuing to improve that over time, though we get very few suggestions or requests these days 🙌🏼

PR #335, opened 2020-09-06
Allow setting topics on posts

Lanny:
### Issue
Does most of #247 

## Description

- Adds infrastructure for topic translations
- Allows users to set the topics for a post when editing or creating a post
- Currently does display topics associated with posts, or allow filtering on topics


robin-macpherson (approved):
All looks good to me! Approving now but need to chat about the roll-out before merging as we've discussed offline 👍🏼 There will also be a few things to resolve with the comment thanks PR presumably going in first

PR #336, opened 2020-09-06
Add "say thanks" to Thread Comments

### LINKED ISSUE #192: FEATURE: Add "thanks" to feedback
#### Perceived Problem

- Currently, there's no way for a Journaler to explicitly indicate that some feedback was useful
- This means we also don't have a great way to show that statistic for people leaving feedback and this could result in people being less motivated to give feedback

- Add a "thanks" ability when receiving feedback that we can then (a) notify the commenter they've been thanked for their helpful feedback, and (b) display stats like "87 thanks received"
- These can even be tied to feedback "types" and we can have special badges like "grammar legend" when you've been thanked for 100 grammar fixes or "dictionary king" for being thanked for 100 alternate word choice suggestions, etc!

#### Screenshots/examples

- Here is an example I saw on the website Lyrics Translate, where they show # of thanks received!
Screen Shot 2020-07-27 at 3 18 16 PM

robin-macpherson:
## Description

**Issue:** closes #192 

Adds the ability to "say thanks" to someone for their feedback, indicating that it was helpful.

## Subtasks

- [x] Write backend portion: mutation, new email notification fn, update object types
- [x] Add UI (like icon)
- [x] Write GQL stuff and regenerate Apollo Hooks
- [x] Make it all work!
- [x] Display correct data in Author Card?


components/Dashboard/Post/PostAuthorCard.tsx
```diff
@@ -55,11 +55,11 @@ const PostAuthorCard: React.FC<PostAuthorCardProps> = ({ author }) => {
       <div className="stats">
         <p className="author-info-heading">Has written</p>
         <ul>
-          <li>10 posts</li>
+          <li>{author.postsWrittenCount} posts</li>
         </ul>
         <p className="author-info-heading">Has received</p>
         <ul>
-          <li>64 thanks</li>
+          <li>{author.thanksReceivedCount} thanks</li>
```
Lanny, line 62: Copy here should use translatable strings

components/Icons/LikeIcon.tsx
```diff
@@ -1,18 +1,32 @@
...
+  title,
+  titleId,
+  filled = false,
+  ...props
+}: React.SVGProps<SVGSVGElement> & SVGRProps) {
   return (
-    <svg width={16} height={14.119} viewBox="0 0 16 14.119" aria-labelledby={titleId} {...props}>
+    <svg
+      width={16}
+      height={14.119}
```
Lanny, line 18: Should have mentioned this when we were looking at this, but at least in the past fractional pixels produced inconsistent rendering and some extra work for the browser that you almost never need. The viewbox and anything that's measured in SVG's abstract units can be fractional without any worries, but width/height should be integers. I think 16x14 would be the way to go here

components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -54,6 +62,40 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment })
...
+    },
+  })
+
+  const createNewCommentThanks = () => {
+    createCommentThanks({
+      variables: {
+        commentId: comment.id,
+      },
+    })
+  }
```
Lanny, line 78: This is just a personal style preference, so feel free to ignore it, but I prefer this kind of style:

```suggestion
  const [createCommentThanks] = useCreateCommentThanksMutation()

  const createNewCommentThanks =  async () => {
    await createCommentThanks({
      variables: {
        commentId: comment.id,
      },
    })
    await  onUpdateComment()
  }
```

because it keeps the logic together, as opposed to `onCompleted` being separated from where the thing calling it is.

components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -137,6 +179,16 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment })
           </Button>
         </>
       )}
+      {!canEdit && currentUserId && (
+        <div className="edit-block">
+          <span
```
Lanny, line 184: Semantically this `<span>` should be a `<button>`

graphql/fragments.graphql
```diff
@@ -6,6 +6,8 @@ fragment UserFragment on User {
   bio
   userRole
   profileImage
+  postsWrittenCount
+  thanksReceivedCount
```
Lanny, line 10: I think this should go on another fragment, maybe `UserWithStatsFragment` like:

```
fragment UserWithStatsFragment {
  ...UserFragment
  postsWrittenCount
  thanksReceivedCount
}
```

and use that on the post page so we don't have to spend the cycles/queries getting these numbers in places where we don't use them (e.g. current user, or when fetching post cards)

nexus/thanks.ts
```diff
@@ -0,0 +1,115 @@
...
+            author: true,
+            thread: true,
+          },
+        })
+
+        const post = await ctx.db.post.findOne({
+          where: {
+            id: comment?.thread.postId,
+          },
+        })
```
Lanny, line 45: You can save a query here by getting the post through the thread with nested includes, an example of doing this is:

```
        const thread = await ctx.db.thread.findOne({
          where: { id: args.threadId },
          include: {
            subscriptions: {
              include: {
                user: true,
              },
            },
            post: {
              include: {
                author: true,
              },
            },
          },
        })
```

nexus/thanks.ts
```diff
@@ -0,0 +1,115 @@
...
+  definition(t) {
+    t.field('createCommentThanks', {
+      type: 'CommentThanks',
+      args: {
+        commentId: intArg({ required: true }),
+      },
+      resolve: async (_parent, args, ctx) => {
+        const { userId } = ctx.request
+        if (!userId) {
+          throw new Error('You must be logged in to create threads.')
```
Lanny, line 27: Error message should be updated

nexus/thanks.ts
```diff
@@ -0,0 +1,115 @@
...
+            },
+          })
+
+          if (!currentUser) throw new Error('User not found.')
+
+          const originalCommentThanks = await ctx.db.commentThanks.findOne({
+            where: {
+              id: commentThanksId,
+            },
+          })
```
Lanny, line 101: These two queries can be done in parallel by not awaiting them immediately, and doing `const [currentUser, originalCommentThanks] = await Promise.all([...])`

Lanny (changes requested):
A handful of minor things, splitting the UserFragment up is probably the most important thing, the rest is negotiable.

components/Dashboard/Post/PostAuthorCard.tsx
robin-macpherson: Great point! Added translations for all strings in this component, including singular/plural form for "post(s)"

components/Icons/LikeIcon.tsx
robin-macpherson: Updated everything to match!

components/InlineFeedbackPopover/Comment.tsx
robin-macpherson: I agree! Great suggestion

components/InlineFeedbackPopover/Comment.tsx
robin-macpherson: Would you be ok with this being a wee follow-up? I'd like to come through and also convert all the strings to translations, too. I could do that and maybe create a transparent `buttonVariant` that we can use on all these icons? (the `<span>`s are existing from previous work currently)

robin-macpherson: If so then I can create the issue and work on it as one of my next things

graphql/fragments.graphql
robin-macpherson: Yep I was a bit unsure of this and had it in my notes to ask your opinion about haha, great suggestion. The only trouble here is that we're using the `AuthorWithLanguagesFragment` in the `PostAuthorCard` that you're referencing. I'm actually a bit unsure what to do there... take a look at my new changes in [this commit](https://github.com/Journaly/journaly/pull/336/commits/f47f94553a49f57426ec9b15364e38385bf07ba8) and let me know you're thoughts

nexus/thanks.ts
robin-macpherson: Done!

nexus/thanks.ts
robin-macpherson: Done! [Like so?](https://github.com/Journaly/journaly/pull/336/commits/3b80cea4943239fe87ff5984c1a6e707b7ddb0f7)

nexus/thanks.ts
robin-macpherson: Done! [Like so?](https://github.com/Journaly/journaly/pull/336/commits/3b80cea4943239fe87ff5984c1a6e707b7ddb0f7) 

(same commit as the previous comment)

graphql/fragments.graphql
Lanny: Hmm, I see, we do have a certain proliferation of fragments. I'm ok with this approach, we might consider taking an approach of each query having its own fragment, so `PostByIdFragment` could combine "AuthorWithLanguagesFragment" with the two stats. Or maybe an "AuthorCardFragment". I think the normal graphql approach would be to have a "AuthorFragment" and anything extra you need just goes in the query, but we need extra fragments because that gives us a named type. Something to think about, but as long as we're not pulling `postsWrittenCount` and `thanksReceivedCount` on like the feed and other places we're using `PostCard` it's good with me.

Lanny (approved):
Dope!

Issue #337, opened 2020-09-09
MY FEED: Add a greeting in the user's feed in a language they're studying? (idea)


robin-macpherson:
#### Idea

- What if we put a nice little greeting in the user's feed that simply says, "Hello Ryan" in one of the languages that they're learning?
- It feels like a very simple overall thing but that might contribute to this notion of Journaly getting to know the user and being personalized to them.

- We could simply get translations in all the languages we have for the word "Hello" and then use that
- If a user is learning multiple languages, we could pick one at random each time they load the feed
- We could also add a time-relevant greeting, such as "good morning" in morning hours and "good afternoon" in afternoon hours. This could be a second iteration

#### Inspiration

I got the idea from studying on Babbel:

![image](https://user-images.githubusercontent.com/34203886/92619220-1627ef00-f276-11ea-9d3b-7d920d62cdfd.png)

PR #338, opened 2020-09-15
General greeting on My Feed page, along with "clear filters" and adding lots of translations

robin-macpherson:
## Description

**Issue:** #337 

- Adds a lavly greeting to the *My Feed* page!
  - This greeting should be translated into the language they are learning
  - If learning multiple languages, it should select one at random on each page load
- Adds a "clear filters" option underneath the language filters
- Adds a bunch of translations to help us tighten that up a good bit
- Makes it clear that "log in" and "sign up" links in the respective forms are links (requested on our Alpha feedback spreadsheet so threw that in here)
- Creates two new `ButtonVariant`s:
  - `Icon`
  - `Link`

## Subtasks

- [x] Add the general greeting in English
- [x] Figure out how to force a specific translation of one string? (@Lanny 🙏🏼)
- [x] Add translations of "hello" for all languages currently being used and as many as possible for now
- [x] Add "clear filters"
- [x] Add lots of translations

## Da Flow

greeting-and-filter-actions-preview


robin-macpherson:
@Lanny I made some good progress on a few small things here! One thing I realized upon digging into the nice little greeting feature is that I'm not entirely sure how to force a specific language translation on a single string on a page. Then we need a mechanism to randomly generate an index for the `languagesLearning` list if the length is more than 1 and then use the translation for the lang at that index.

Getting a bit late here so haven't looked too hard yet, but thought I'd bring you in on it, too! 😄

components/Dashboard/Post/PostComment.tsx
```diff
@@ -79,7 +84,9 @@ const PostComment: React.FC<PostCommentProps> = ({
                 ? `${comment.author.name} (@${comment.author.handle})`
                 : `@${comment.author.handle}`}
             </span>
-            <span className="comment-date">{formatDateRelativeToNow(comment.createdAt)} ago</span>
+            <span className="comment-date">
+              {formatDateRelativeToNow(comment.createdAt)} {t('relativeTimeWord')}
+            </span>
```
robin-macpherson, line 89: @Lanny I realized here that currently our little time library function is what returns "minutes" or "hours" or "days", etc... for the relative time comparison, and the position of the equivalent of "ago" will change based on the language 🧐

components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -91,7 +110,9 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
       <Head>
         <title>{pageTitle}</title>
       </Head>
-      <h1>My Feed</h1>
+      <h1>
+        {greetings[greetingLanguage]} {currentUser.name || currentUser.handle}!
```
robin-macpherson, line 114: @Lanny hmm, TS shouts at me here because of an implicit `any` type, even though it seems to clearly know the type... 🤔 Tried a few annotations but I seem to be doing something wrong

image

robin-macpherson:
Ah, also, I wonder what to do about right-to-left languages 😬

components/Dashboard/MyFeed/MyFeed.tsx
robin-macpherson: Fixed this now 🙂 ✅

robin-macpherson:
> Ah, also, I wonder what to do about right-to-left languages 😬

I wrote a little something to try and handle this. Not sure if it's the most elegant solution but it works! 😄

robin-macpherson:
@Lanny okie dokie so everything seems to be working perfectly locally, but I noticed while testing in the staging environment that the greeting feature is weird due to multiple re-renders occurring on each page load?

robin-macpherson:
@Lanny since we're experiencing some tricky times with you having power outages, I'm going to go ahead and merge this in. I want to make sure we keep the momentum going with our alpha users and that they see updates, and I feel quite confident that I've tested it well and everything works. We can happily do a follow-up to address any comments you have once you get back online! 😄

PR #343, opened 2020-09-24
Journalers can follow each other!

### LINKED ISSUE #324: FEATURE: Enable users to "Follow" other users
#### Perceived Problem

- As our user base grows, it's going to become more and more difficult to know when your favorite Journalers have written a post you might want to read.

- Enable the ability to "follow" another user!

#### Iterations
1. A simple email notification
2. Show posts from users you follow first (?)
3. Look at ways to mix the results in a way that is weighted towards people you follow, but still shows posts that need feedback and that are a good match
4. Use ML to make this more sophisticated

robin-macpherson:
## Description

**Issue:** closes #324 

This PR builds out the ability for Journalers to follow each other with one initial entry point.

## Subtasks

- [x] Update schema
- [x] Migration
- [x] Write new mutations
- [x] Get follow & unfollow working from `PostAuthorCard`
- [x] Add new filter action for posts from followed users
- [ ] ~Wire up follow notification email (?)~


components/Dashboard/Post/PostAuthorCard.tsx
```diff
@@ -26,6 +32,37 @@ const PostAuthorCard: React.FC<PostAuthorCardProps> = ({ author }) => {
...
+    followUserMutation({
+      variables: {
+        followedUserId: author.id,
+      },
+    })
+  }
+
+  const [unfollowUserMutation, { loading: unfollowLoading }] = useUnfollowUserMutation({
+    onCompleted: () => {
+      refetch()
```
robin-macpherson, line 54: This is currently not working as expected, and I'm noticing some flickering upon clicking this but also just page refresh

graphql/fragments.graphql
```diff
@@ -28,6 +34,12 @@ fragment UserWithLanguagesFragment on User {
...
   }
+  following {
+    id
+  }
+  followedBy {
+    id
+  }
 }
 
 fragment AuthorFragment on User {
```
robin-macpherson, line 45: Still feeling quite unsure about how the fragments are evolving and where to put these. I originally opted to put them in `UserWithStatsFragment` (wasn't too sure about that either, but it sort of made sense and we're already using that for the `author` in the `PostAuthorCard`) but then realized I needed `currentUser` to get `currentUser.following` and that uses `UserWithLanguages...`. 

Is there a point where it makes sense to have a separate query? Either separate `currentUser` queries, or simply `followersQuery` for example? Or do you think we're fine doing it like this?

components/Dashboard/Post/PostAuthorCard.tsx
```diff
@@ -3,18 +3,24 @@ import Link from 'next/link'
...
+  useUnfollowUserMutation,
 } from '../../../generated/graphql'
 import { useTranslation } from '../../../config/i18n'
 import BlankAvatarIcon from '../../Icons/BlankAvatarIcon'
 import { languageNameWithDialect } from '../../../utils/languages'
+import Button, { ButtonVariant } from '../../../elements/Button'
 
 type PostAuthorCardProps = {
   author: Author | any
+  currentUser: UserType | any
```
robin-macpherson, line 19: Would love to handle this better

robin-macpherson:
@Lanny still WIP but felt like a good time for an early review, left a few comments 🙂

robin-macpherson:
@Lanny just as a status update, I just added the new filter logic to the `feed` query FE & BE and also wrote the new filter action and got that working, albeit with a couple of TS errors

components/Dashboard/Post/PostAuthorCard.tsx
```diff
@@ -39,6 +66,15 @@ const PostAuthorCard: React.FC<PostAuthorCardProps> = ({ author }) => {
           </a>
         </Link>
         <p className="author-name">{author.name || author.handle}</p>
+        {currentUser && currentUser.id !== author.id && (
+          <Button
+            variant={ButtonVariant.Link}
+            onClick={hasFollowedAuthor ? handleUnfollowUser : handleFollowUser}
+            loading={followLoading || unfollowLoading}
+          >
+            {hasFollowedAuthor ? 'Unfollow' : 'Follow'}
```
Lanny, line 75: Little thing but ideally these would be translated

Lanny (approved):
Two small things, but looks great!

components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -155,12 +160,25 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
             </Button>
             <Button
               variant={ButtonVariant.Link}
+              className={`filter-action-btn ${
+                JSON.stringify(selectedLanguageFilters) ===
+                JSON.stringify(Array.from(userLanguages.values()))
+                  ? 'active'
+                  : ''
```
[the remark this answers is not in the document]
robin-macpherson: Yep yep I knew that was not the most elegant piece of code I'd ever written 😂 Ok so I just improved it quite a bit I think and we can revisit this in a follow-up.

I finally just added `lodash` since they have a nice `_.isEqual()` method that performs a deep comparison between two arrays to determine if they are equivalent (also works for comparisons between objects, dates, strings, array buffers, and more). This will also cover the issue of ordering.

So I used `_.isEqual()` to create that piece of state you suggested and then cleaned up the ternary. Does that feel like a decent improvement for the time being? (I'll merge it in but can happily revisit ASAP if you don't like this solution)

components/Dashboard/Post/PostAuthorCard.tsx
robin-macpherson: Great catch! Done!

PR #347, opened 2020-09-29
Fix filter pagination bug

### LINKED ISSUE #346: 🐞 When adding filters while on page 2/3/4... the page doesn't reset
#### Description

I noticed that if you are on page 4 of the results and then select a language, you will remain on page 4 of the results. This also means that if there are less than 4 pages of results, you will see "nothing to see here yet..."

robin-macpherson:
## Description

**Issue:** fixes #346 

I noticed that if you are on page 4 of the results and then select a language, you will remain on page 4 of the results. This also means that if there are less than 4 pages of results, you will see "nothing to see here yet..."

This PR tries to push the user back to page 1 whenever they adjust the filters. @Lanny does this approach make sense to you, and how I implemented it?

## Subtasks

- [x] Fix the thing

### DIFF
--- a/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/components/Dashboard/MyFeed/MyFeed.tsx
@@ -17,7 +17,7 @@ import Button, { ButtonVariant } from '../../../elements/Button'
 import { greetings } from './greetings'
 import useToggle from '../../../hooks/useToggle'
 
-const NUM_POSTS_PER_PAGE = 9
+const NUM_POSTS_PER_PAGE = 2
 
 type Props = {
   currentUser: UserType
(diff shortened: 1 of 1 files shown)


robin-macpherson:
Ah, never mind. This approach doesn't work (needed to test it on staging to have enough post data/pagination). Will try another approach in the morning

robin-macpherson:
@Lanny also tried refactoring to use local state which might be a little cleaner anyway, but that doesn't seem to work either. I guess we can just push them directly via the router but I'm a little confused why this doesn't work? Shouldn't the `useFeedQuery` refetch if the value of one of the variables is updated?

Lanny:
`useFeedQuery` should refetch when variables change. We should be pushing changes to the router though since page number is part of the URL. Not sure why this isn't working. I'd expect it to bounce you back to page 1 but desync the displayed page and URL

xferguson:
I opened an alternate approach that involves pushing to the router. I am not convinced that we want to use local state for this one just because the router controls a whole slew of variables, and we would end up with cross-cutting concerns. Since this has been sitting for a month, I figured an alternative may be welcome 😄

Issue #352, opened 2020-10-27
Add the notion of a "Language Level"


robin-macpherson:
#### Perceived Problem

- Currently there is no way to indicate what level a user has in a language
- This also means we can't indicate to potential readers the level of a post author
- Users have asked for this feature a lot
- We've also identified a big improvement that could be made to the DB design with a unified `LanguageRelation` table (and removing `LanguageLearning` & `LanguageNative`

#### Ideas / Proposed Solution(s)

- Create new `LanguageRelation` table and refactor all code accordingly
- Add this feature so users can select their level when adding a language, and display this on post cards in the feed and in the post itself

PR #353, opened 2020-10-27
Create new `LanguageRelation` table & refactor all existing code

robin-macpherson:
## Description

**Issue:** #352 (part 1 of 2)

This is PR 1/2 for adding language levels to the app.
This PR creates the new unified `LanguageRelation` table and refactors all code without any changes to the UI or experience.
Defaults `LanguageLevel` to `BEGINNER`.

## Subtasks

- [x] Datamodel changes
- [x] refactor BE code
- [x] refactor gql files and re-gen Apollo hooks
- [x] refactor FE code accordingly and get build passing
- [x] go over the flows, make sure everything still works, and make any decisions on unifying forms, etc.


components/Modals/WelcomeModal/WelcomeModal.tsx
```diff
@@ -20,7 +20,7 @@ const WelcomeModal: React.FC = () => {
 
   useEffect(() => {
     const hasChosenLanguages =
-      data?.currentUser?.languagesNative.length && data?.currentUser?.languagesLearning.length
+      data?.currentUser?.languages.length
```
Lanny, line 23: We need to make sure there's at least one native and one learning language here. A sorta tricky way to do it:

```suggestion
      (data?.currentUser?.languages.length || [])
        .reduce((acc, { level }) => acc | (level === LanguageLevel.Native ? 1 : 2), 0) === 3
```

components/PostEditor/PostEditor.tsx
```diff
@@ -88,8 +88,8 @@ const PostEditor: React.FC<PostEditorProps> = ({
     debounceTime: 1000,
   })
 
-  const { languagesLearning = [], languagesNative = [] } = currentUser || {}
-  const userLanguages = [...languagesLearning, ...languagesNative].map(({ language }) => {
+  const { languages = [] } = currentUser || {}
+  const userLanguages = [...languages].map(({ language }) => {
```
Lanny, line 92: No need to clone the array here

```suggestion
  const userLanguages = languages.map(({ language }) => {
```

graphql/fragments.graphql
```diff
@@ -16,19 +16,13 @@ fragment UserWithStatsFragment on User {
 
 fragment UserWithLanguagesFragment on User {
   ...UserFragment
-  languagesLearning {
+  languages {
     id
```
Lanny, line 20:
```suggestion
    id
    level
```

nexus/language.ts
```diff
@@ -155,9 +76,72 @@ schema.extendType({
...
+        }
+
+        const relation = await ctx.db.languageRelation.findOne({
+          where: {
+            userId_languageId: {
+              languageId: args.languageId,
+              userId,
+            },
+          },
+        })
```
Lanny, line 137: Would this work?

```suggestion
        const relation = await ctx.db.languageRelation.findOne(relFilter)
```

Lanny (commented):
Looks good! Had a handful of comments but I think it's quite close

components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -50,13 +50,9 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
...
-  for (let languageLearning of currentUser.languagesLearning) {
-    if (languageOptionIds.has(languageLearning.language.id))
-      userLanguages.add(languageLearning.language.id)
-  }
-  for (let languageNative of currentUser.languagesNative) {
-    if (languageOptionIds.has(languageNative.language.id))
-      userLanguages.add(languageNative.language.id)
+  for (let language of currentUser.languages) {
+    if (languageOptionIds.has(language.language.id))
+      userLanguages.add(language.language.id)
```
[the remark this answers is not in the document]
robin-macpherson: I agree, this is one of the things I wanted to discuss regarding naming between language, language relation, and user language

components/Modals/WelcomeModal/WelcomeModal.tsx
robin-macpherson: Yeah I agree, I think we could discuss how the end UX/flow should be and how we can wire that up to this

components/PostEditor/PostEditor.tsx
robin-macpherson: Good point!

nexus/language.ts
robin-macpherson: Yep! I think after the refactor I just thought that it made sense to simply do it inside the `where` since `relFilter` was only being used once now

graphql/fragments.graphql
robin-macpherson: Great catch! ✅

nexus/language.ts
robin-macpherson: Yes and no! bahaha yes on a basic level to use the variable instead, but it looks like this is now an invalid `languageRelation.findOne()` invocation o(>.<;)o

robin-macpherson: Weird though because digging around the `index.d.ts` file I still see `userId_languageId` as it was, so maybe something else is causing the error

Lanny (approved):
👍

PR #362, opened 2020-11-03
Fix Edit Button Bug

### LINKED ISSUE #361: Edit post throws an error with i18n
#### Description

If a user has their browser set to a language other than English, clicking the edit post button throws an error.
Using the i18n Next Link component should resolve this.

robin-macpherson:
## Description

**Issue:** fixes #361 

## Subtasks

- [ ] ...

### DIFF
--- a/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/components/Dashboard/MyFeed/MyFeed.tsx
@@ -15,7 +15,6 @@ import {
   useLanguagesQuery,
   LanguageLevel
 } from '../../../generated/graphql'
-// import Select from '../../../elements/Select'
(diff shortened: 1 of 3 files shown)


Lanny:
Sorry for leading you down the wrong path on this one @robin-macpherson , looking at upstream it seems like next-i18next just doesn't support locale paths with dynamic routing. PR to remove dynamic routing here: https://github.com/Journaly/journaly/pull/366

robin-macpherson:
> Sorry for leading you down the wrong path on this one @robin-macpherson , looking at upstream it seems like next-i18next just doesn't support locale paths with dynamic routing. PR to remove dynamic routing here: #366

Not to worry @Lanny ! Thanks for that fix in #366 :)

PR #363, opened 2020-11-03
Add update some lines and add notes for Windows users

### LINKED ISSUE #364: Readme needs updates and options for Windows machines
#### What

- There need to be updates to the documentation for the current setup and for Windows machines.

#### Why

- Attempting to follow the readme to install Journaly on my local Windows machine didn't work.

#### How

- I have already opened a PR servicing some of these changes #363 
- There are two other changes that _may_ be necessary, but it may also be worth splitting them into another issue (or two)

xferguson:
## Notes

I don't seem to be able to add Notes to the Kanban, so I am unable to follow the normal workflow for this PR. I created an issue for this (after the PR) #364 but am unsure how to get it on the Kanban.

## Description

closes #364 

### Changes
These are changes I needed to make to the `Readme.md` file in order to try and get Journaly running locally on my Windows machine.

1. I updated the section Postgres installation to include how one installs it on Windows.
1. I fixed what I believe as a typo, referencing the `backend` directory when it was supposed to be the `prisma` directory.
1. I made additions to where we have to make changes to `.env.example`, as there are two instances of that file. This may be incorrect, and it is working for me now, but I'm unsure if I have superfluous files at the moment (I am not familiar with Prisma yet).

### Items left unchanged
1. `npm run reseed-db` will not work on Windows because `source` (called by `empty-db`) isn't available. I was able to run the sub-scripts manually, but am unsure the best way to compile that into a new script. I'd recommend making another issue to handle that.
1. The bottom of the readme references `localhost:3000/api/playground` for GraphQL. When I load that I get the error that I cannot connect to the server. This may be due to changes in configuration, or due to something I didn't replicate properly in my setup, so I didn't want to change it. I did want to get someone's eyes on it who is familiar with the setup to make sure if it needs to be updated.

## PR Boilerplate
**Issue:** #364 

- [ ] I have added this PR to the Journaly Kanban project ✅ <-- I am unable to do this

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,16 @@ The data model/DB schema lives in `prisma/schema`, server side code and some uti
 
 #### Setting Up Your Local DB Instance
 
-1. Install Postgres with Homebrew
+1. Install Postgres
+- Mac/Linux users should use Homebrew
    _Note that this set your Postgres DB to run when your computer starts up and will stay running in the background_.
 
    ```bash
    $ brew install postgres
    $ brew services start postgresql
    ```
+
+- Windows users should download and install Postgres from [postgresql.org](https://www.postgresql.org/download/windows/).
 
 1. Start up your Postgres shell and create your user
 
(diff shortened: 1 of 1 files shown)


robin-macpherson (approved):
Awesome, thanks @xferguson ! The README definitely needs some love, this is very helpful 😊

PR #365, opened 2020-11-04
call pagination reset from filters

### LINKED ISSUE #346: 🐞 When adding filters while on page 2/3/4... the page doesn't reset

xferguson:
## Description

**Issue:** closes #346 

I don't mean to compet with @robin-macpherson's solution in #347 but it seems like that has been hanging for a while, and a lot of issues appear to be assigned but maybe on hold. So I grabbed what appeared to be something I could tackle

Essentially, my fix creates a new function to reset the pagination when the filters are edited. I use the [router.push()](https://nextjs.org/docs/api-reference/next/router#routerpush) method to adhere to single responsibility principles and ensuring that all intended "side effects" are taken care of (such as the url updating properly).

I would love someone to look this over and give me feedback. There are a few items in particular I was curious about (being new to the codebase) and would love some eyes on:

1. I am currently using a spread notation to rebuild the `router` object because I am unsure how you all prefer to handle immutability. I see there is `lodash` installed, so I might want to take a swing at using their `cloneDeep` or `omit` since this isn't a resource intensive context.
1. I wasn't sure if there was a reasonable `util/` folder, or similar place, to put this `resetPagination` function. Is there any reason for this to be called outside of this component? If that's not a concern right now, then maybe the question is moot. I just wanted to check.

## Subtasks

- [x] Chosen immutability method for `router` has been approved for style ✅
- [x] Verified the location off the `resetPagination` function is good ✅

### DIFF
--- a/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/components/Dashboard/MyFeed/MyFeed.tsx
@@ -23,7 +23,7 @@ import { greetings } from './greetings'
 import useToggle from '../../../hooks/useToggle'
 import Select from '../../../elements/Select'
 
-const NUM_POSTS_PER_PAGE = 9
+const NUM_POSTS_PER_PAGE = 2
 
 type Props = {
   currentUser: UserType
@@ -76,8 +76,8 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
    * Pagination handling
    */
   // Pull query params off the router instance
-  const { query } = useRouter()
-  const currentPage = query.page ? Math.max(1, parseInt(query.page as string, 10)) : 1
+  const router = useRouter()
+  const currentPage = router.query.page ? Math.max(1, parseInt(router.query.page as string, 10)) : 1
 
   // fetch posts for the feed!
   const { loading, error, data } = useFeedQuery({
@@ -99,6 +99,17 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
   if (currentUser.languages.length === 1) {
     greetingLanguage = currentUser.languages[0].language.name
   }
+
+  const resetPagination = () => {
+    // filter out page key to reset the url
+    router.push({...router, query: Object.keys(router.query).reduce((newQuery, key) => {
+      if (key !== 'page') {
+        newQuery[key] = router.query[key]
+      }
+      return newQuery
+    }, {})})
+  }
+
 
   const learningLanguages = currentUser.languages.filter(({ level }) => level !== LanguageLevel.Native)
   if (learningLanguages.length > 0) {
@@ -152,12 +163,16 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
             options={languageOptions}
             selectedOptionValues={selectedLanguageFilters}
             placeholder="Languages"
-            onAdd={(id) => setSelectedLanguageFilters([...selectedLanguageFilters, id])}
-            onRemove={(id) =>
+            onAdd={(id) => {
+              setSelectedLanguageFilters([...selectedLanguageFilters, id])
+              resetPagination()
+            }}
+            onRemove={(id) => {
               setSelectedLanguageFilters(
-                selectedLanguageFilters.filter((languageId) => languageId !== id),
+                selectedLanguageFilters.filter((languageId) => languageId !== id)
               )
-            }
+              resetPagination()
+            }}
           />
           <div className="filter-actions">
             <Button
--- a/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/components/Dashboard/MyFeed/MyFeed.tsx
@@ -23,7 +23,7 @@ import { greetings } from './greetings'
 import useToggle from '../../../hooks/useToggle'
 import Select from '../../../elements/Select'
 
-const NUM_POSTS_PER_PAGE = 2
+const NUM_POSTS_PER_PAGE = 9
 
 type Props = {
   currentUser: UserType
--- a/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/components/Dashboard/MyFeed/MyFeed.tsx
@@ -102,12 +102,9 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
 
   const resetPagination = () => {
     // filter out page key to reset the url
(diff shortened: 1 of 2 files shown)


xferguson:
I _attempted_ to run `npm run format` on this, but got an error. I suppose if I switch to VSCode instead of my trusty, dusty SublimeText and add in a prettier extension, this won't be an issue. So I can look at that as well. Probably not today though -- bed time.

robin-macpherson:
Thanks a lot for this PR @xferguson ! Yeah I was working on a fix for this but unfortunately a bunch of high priority stuff came up and I didn't get back to my PR yet. I'm perfectly happy to take yours instead and I think it's a good solution!

Regarding question 1, we're fine with spread notation here - especially for the time being - but that's a great thought regarding `lodash`.

As for question 2, I think let's leave this in the current file for the time being. I think this could definitely end up being used in some other places, such as the J-Forum when we build that and also perhaps when we build out direct messaging & inbox soon. But the approach we have taken so far is to generally keep things next to where they are used until we actually arrive at a point where something needs to be reused and then break things out at that point.

Everything looks great to me other than it looks like you have one small TypeScript error. Usually I always run `npm run build` once I'm done with a chunk of work and that will throw any TS errors you have. Let me know if you need any help with that!

Happy to merge this in once we fix those errors 🙂

components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -100,6 +100,17 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
     greetingLanguage = currentUser.languages[0].language.name
   }
 
+  const resetPagination = () => {
+    // filter out page key to reset the url
+    router.push({...router, query: Object.keys(router.query).reduce((newQuery, key) => {
```
Lanny, line 105:
```suggestion
    // filter out page key to reset the url
    const newQuery = {...query}
    del newQuery.page
    router.push({...router, query: newQuery })
```

Would work as well, seems a little more direct to me, but either way is good in my book

Lanny (commented):
Thanks for the PR! Like Robin said, just one issue with the typechecker not being happy, left a suggestion that I think would satisfy it. Type annotations would work as well

components/Dashboard/MyFeed/MyFeed.tsx
xferguson: @robin-macpherson + @Lanny -- Thanks for the notes!

I'll go ahead and add in Lanny's suggestion -- new to typescript, so I wouldn't have caught in sublime. I'll also make a point to run `npm run build` in the future -- Just assumed that was done for me by git hooks. Is there travis or some other ci hooked up to this repo?

I'll make a note about `npm run build` in the [contributing-guide.md ](https://github.com/Journaly/journaly/blob/master/docs/contributing-guide.md) and add it to this PR as well.

xferguson:
I pushedchanges to the function and to the contribution docs.

It turns out I can't run `npm run build` on my machine because [Nexus@0.25.0 doesn't support Windows CLI](https://github.com/prisma-labs/graphql-framework-experiment/issues/965#issuecomment-645485119). They [released Windows support](https://github.com/prisma-labs/graphql-framework-experiment/pull/1184) in July, and we could upgrade to the [Latest (0.27.0)](https://github.com/prisma-labs/graphql-framework-experiment/compare/0.25.0...0.27.0) which should work.

I'd be glad to take on that work, but wouldn't recommend blocking this fix with the upgrade because even though it's a minor upgrade, I'm not at all familiar with Nexus and wouldn't know how/what to properly test.

Windows sure is a hassle. This is why I use VSCode on a Mac for work. alas, no Mac for personal stuff 👎  -- Regardless, getting everything operable for Windows is good for contribution.

xferguson:
Is the error you are getting in the Vercel output? I can't access that at the moment, it seems. I created a Vercel account, but when I click on the Details link, I get a 404.

Sorry if this is more of a hassle than a help. Consider it QA Engineering on the contribution process 😉

xferguson:
Thanks, @Lanny -- embarrassingly simple copy-paste mistaje. I need to get vsc on my home laptop before I take anything else on this project so it will yell at me if I use undeclared variables.

Lanny:
>Sorry if this is more of a hassle than a help. Consider it QA Engineering on the contribution process 😉

Haha, definitely a help, fixing this has been in limbo for a while so it's great to see someone pick it up. And you're our first external contributor so we really appreciate seeing where the pain points are. We should get something like Travis going for sure, Vercel builds the project so that's what we've been leaning on, but it also does a deploy using AWS credentials so it's restricted to core contributors for security reasons. Getting CI that will run for people outside of core is definitely something we should get going.

Re: nexus upgrade, yeah, if you want to upgrade to 0.27 and test it out on windows we'd definitely accept a PR for that. Nexus-the-framework has unfortunately stopped development at this point (I imagine 0.27 is the final release) so we need to get off of it ("nexus-schema" which nexus-the-framework wraps, plus some glue code, is the eventual go-forward) but just upgrading to the last nexus framework version would be a good move short term. Hopefully it's as simple as bumping the version number in package.json, I don't expect any serious breakage in that upgrade. If you run into any issues getting it set up on windows feel free to reach out, I'm sure there's some things (e.g. like you pointed out reseed-db expects `source` it be a command) which should be pretty straightforward to fix but we've just never stumbled across since we've never even tried running in a windows environment

Lanny (approved):

robin-macpherson (approved):
Awesome thanks again for this @xferguson! As @Lanny said it's really cool to have you be our first external contributor! Sorry that you've had a bit of trouble running some of the scripts. We'll definitely be happy to do what we can to support getting you up and running there.

Going to merge this in for now 🎉

robin-macpherson:
@xferguson just wanted to drop another thanks here for helping on this PR 🎈 @Lanny and I chatted and thought of a few more really good issues that would be great first issues just in case you are looking for more things to grab!

I just created a few new issues for them with the "good first issue" tag in case you would like to take a look 🙂 I did my best to also quickly throw some helpful background/context in those issues and we'll be more than happy to help with anything else if you decide to pick something up!

PR #366, opened 2020-11-04
Remove locale subpaths

Lanny:
## Description

Fixes issue with users with a german browser setting not being able to edit posts. Does this by disabling locale subpaths functionality with next-i18next (dynamic routes and locale subpaths don't work together, apparently. I liked locale subpaths, so I'm sad to see them go, but if they don't work then they just don't work.

Also added a little widget to the profile page to let you change languages for testing purposes. Probably doesn't belong here long term, but it seemed alright for now.

## Screenshots

Screen Shot 2020-11-04 at 4 45 32 PM

### DIFF
--- a/components/Icons/FlagDEIcon.tsx
+++ b/components/Icons/FlagDEIcon.tsx
@@ -1,0 +1,18 @@
+import * as React from 'react'
+interface SVGRProps {
+  title?: string;
+  titleId?: string;
+}
+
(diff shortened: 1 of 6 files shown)


robin-macpherson (approved):
Wooo awesome!!! Nice idea with the flags and UI language select functionality. Would be good to discuss that one more but super excited to merge this one in and have that bug fixed!

PR #376, opened 2020-11-09
Badges v1

Lanny:
## Description

Adds badges, woo! Badges are visible on the profile pages.

Current badges are:

- Alpha User (folks who made at least one post during the alpha period)
- Beta User (everyone else, plus all new signups until we're done with beta)
- 10 Posts
- 100 Posts
- Code Contributor

I'd like to add badges to author cards as well, but right now you can have up to 4 badges, and I'm sure we're going to add plenty more so I think it would be a bit clutter-y there. I'd like to implement "badge pinning" so users can select 1~3 badges that get displayed outside of the profile, but that's for a different PR.

## Migrations

Order of operations here is going to be

1. Run `migrate:up` against prod
2. Deploy code
3. Run backfill for alpha/beta users
4. Run manual code contrib backfills

See `prisma/migrations/20201101185157-add-badges-table/README.md` for the two queries to backfill alpha/beta badges.

Run this query to add the code contributor badge:

```sql
INSERT INTO "UserBadge" ("userId", type) VALUES (<USER ID HERE>, 'CODE_CONTRIBUTOR');
```

Our ids are 9 and 25. @xferguson @Demezis y'all can post your profile URLs in this PR if you'd like the contributor badge as well.

## Screenshots

Screen Shot 2020-11-08 at 7 03 54 PM


robin-macpherson (commented):
Woooooo so happeh to see this 😄 I know this was already a ton of work so definitely open to discussion on the timing of things, but I feel like we should notify users somehow of reaching each milestone. Should we potentially set up a new email for badge/milestone accomplishment?

Lanny:
Yeah, agree some kind of notification would be a good idea. In-app notifications would be ideal for this use case I think but e-mails would be a good stop-gap for the time being.

Demezis:
Nice badges! I'd like to have the contributor badge 😄 my id is `330`

robin-macpherson (approved):
WOOOO 🦡 🚀

robin-macpherson:
> Nice badges! I'd like to have the contributor badge 😄 my id is `330`

Woohoo! You should see the badge on your profile now 😄 Also @xferguson I grabbed your ID and you should see the badge too 🙌🏼

PR #378, opened 2020-11-09
Post page SSR improvements

Lanny:
## Description

Prior to this PR, SSR rendered posts (and apparently profile pages) return the loading spinner and nothing else, even though the Apollo cache is populated. If your browser is reasonably fast then this is a nearly unnoticeable impact , BUT it means crawlers can't get any of our excellent content. This seems to trace back, at least in part, to a bug in one of our apollo deps which is resolved by upgrading, as evidenced by this fixing the profile page despite no changes to that page.

As part of the fix, I consolidated the queries for current user and post data on the post page, meaning on load we have to hit the graphql endpoint one less time, which is a small but still significant improvement in load time/resources.

Going forward, I think we should try to make no more than one query per top level page, to save on roundtrips. Pages that need to make a top level query should keep that query in `graphql/pages/<page name>Page.graphql`, and bundle together any fields needed to render the page there. This doesn't mean sub-components can't make their own queries (e.g. nav will have its own call to get the current user), just that pages themselves shouldn't be using multiple queries.


robin-macpherson (approved):
Woo this is great! I know we just got some small merge conflicts from the other PR but approving in advance ✅

PR #379, opened 2020-11-09
Enable post deletion

### LINKED ISSUE #369: Enable Post Deletion
#### Perceived Problem

- Currently users cannot delete published or draft posts
- This is a little more complicated than it should be because Prisma does not yet have cascading deletes, so we need to manually handle that in our mutation

- Let's wire up post deletion

#### Helpful background/context

- We have existing patterns for delete operations, such as comments
- See how we handle deleting `Thread`s along with their existing `Subscription`s for an example of handling the cascading issue
- See our `Button` component and usages for examples of that

Demezis:
## Description

**Issue:**: closes #369 

## Subtasks

- [x] Add delete post mutation
- [x] Add delete button and change button style
- [x] Add a delete confirmation
- [x] redirect and refetch

## Screenshots
Screenshot 2020-11-09 173636
image


robin-macpherson:
This is looking great @Demezis !

public/static/locales/en/post.json
```diff
@@ -7,12 +7,20 @@
...
   "languageLabel": "Language",
   "save": "Save",
   "finishAction": "Finish post",
   "noLanguageError": "Please select a language.",
   "emptyTitleError": "Please specify a non-empty title.",
   "draftPublishedSuccess": "Post successfully published!",
-  "postBy": "by"
+  "postBy": "by",
+  "deleteModal": {
+    "title": "Do you want to delete the post?",
```
robin-macpherson, line 22:
```suggestion
    "title": "Are you sure you want to delete this post?",
```

elements/Button/Button.tsx
```diff
@@ -153,11 +153,12 @@ const Button: React.FC<Props> = (props) => {
...
           background-color: ${theme.colors.white};
-          border: 1px solid ${theme.colors.white};
+          border: 1px solid ${theme.colors.red};
         }
         .${ButtonVariant.DestructiveSecondary}:hover:not(:disabled) {
-          background-color: gray;
-          border: 1px solid gray;
+          color: ${theme.colors.white};
+          background-color: ${theme.colors.red};
+          border: 1px solid ${theme.colors.white};
```
robin-macpherson, line 161: Thanks for fixing these! 🙂🙏🏼

public/static/locales/en/post.json
Demezis: Yes, I know about this, I have fixed it, but didn't pushed it
`
  "deleteModal": {
    "title": "Delete Post",
    "body": "Are you sure you want to delete this post?",
    "delete": "Delete",
    "cancel": "Cancel"
  }
`

robin-macpherson: Awesome, looks good!

Demezis:
> I'm wondering if we should make a more generic `ConfirmationModal` rather than a specific modal for each confirmation action?
You are right, I will make the modal more generic, there is no particular reason to make a specific modal for each confirmation.

robin-macpherson:
> > I'm wondering if we should make a more generic `ConfirmationModal` rather than a specific modal for each confirmation action?
> 
> You are right, I will make the modal more generic, there is no particular reason to make a specific modal for each confirmation.

Awesome, thanks!

Demezis:
I don't know how to properly update data after mutation, so please check it.

nexus/post.ts
```diff
@@ -424,5 +424,119 @@ schema.extendType({
...
+        })
+
+        if (!currentUser) {
+          throw new Error('User not found.')
+        }
+
+        hasAuthorPermissions(post, currentUser)
+
+        const deleteFirstPhasePromises = [
+          ctx.db.comment.deleteMany({
```
Lanny, line 463: I think "comment thanks" (that is thanks given to comments within threads) need to be deleted before you can delete comments, unfortunately. It's lousy that we don't have cascading deletes.

Lanny (approved):
Pulled it down locally and everything looks great. I'm good to merge if you are @robin-macpherson

robin-macpherson (approved):
Looks great! Thanks again for this awesome work @Demezis 🙂

PR #381, opened 2020-11-10
Create Internal DB Client Package

### LINKED ISSUE #374: Create internal package for Prisma so that it can be imported by multiple services
#### Perceived Problem

- We need to build additional micro-services such as for email batching that will need to talk to the database
- It would be best to do this through Prisma Client, but currently this can't be imported from those micro-services.

- Create an internal "package" for our generated Prisma Client so that we can instantiate it in the email batching service and use that to talk to the database

robin-macpherson:
## Description

**Issue:** closes #374

This PR creates an internal database client as a package that we can import into multiple micro-services and access the DB via our custom-generated Prisma Client.

Immediate use cases are:
1. Email batching with J-Mail
2. Inline images in posts

The new package is called `j-db-client` which now contains all database-related scripts.

**Commands:**
- Migrations:
- Re-generate Prisma Client:

## Subtasks

## Migrations

There are no migrations to apply.


Lanny (approved):

robin-macpherson:
@xferguson & @Demezis just to make you both aware of this change. I'm working on updating documentation and should hopefully finish that by the end of the weekend

PR #384, opened 2020-11-12
Include PostComments in commentCount on PostCards

### LINKED ISSUE #383: Include post comments in the comment count on Post Cards (?)
#### Perceived Problem

- I'm noticing that when you first look at the feed, it appears that a lot of posts don't have comments due to the 0 comment count shown on the `PostCard`
- However, once clicking into a bunch of them, I see there are post-level comments
- I think this might make it appear that there is less engagement than there actually is

- Since the main purpose of displaying that comment count on the post card is to indicate post engagement, and we do want people to have a good impression that it is worth it to write posts, I wonder if we should also include `PostComment`s in the count?

robin-macpherson:
## Description

**Issue:** closes #383 

I've been noticing that when you first look at the feed, it appears that a lot of posts don't have comments due to the 0 comment count shown on the PostCard, which gives a perception that a lot of posts don't have engagement.
However, once clicking into a bunch of them, I see there are post-level comments.

Since the main purpose of displaying that comment count on the post card is to indicate the level of post engagement, and we do want people to have a good impression that it is worth it to write posts, I think this might help achieve a more accurate perception.

## Subtasks

- [x] Update `commentCount` field

## Screenshots

**Left:** Updated count (shows 5), **Right:** Old count (shows 0)

image

**Here you can see that basically all counts are higher with this update including two that were 0 before**

image

### DIFF
--- a/nexus/post.ts
+++ b/nexus/post.ts
@@ -104,18 +104,20 @@ schema.objectType({
     t.model.publishedAt()
     t.int('commentCount', {
       resolve: async (parent, _args, ctx, _info) => {
-        const threadCommentCount = await ctx.db.comment.count({
-          where: {
-            thread: {
-              postId: parent.id,
+        const [threadCommentCount, postCommentCount] = await Promise.all([
+          ctx.db.comment.count({
+            where: {
+              thread: {
+                postId: parent.id,
+              },
             },
-          },
-        })
-        const postCommentCount = await ctx.db.postComment.count({
-          where: {
-            postId: parent.id
-          }
-        })
+          }),
+          ctx.db.postComment.count({
+            where: {
+              postId: parent.id
+            }
+          })
+        ])
         return threadCommentCount + postCommentCount
       },
     })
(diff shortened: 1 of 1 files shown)


robin-macpherson:
@Lanny I decided to just throw a quick PR together for this one, and looking at it I feel like it really does help. The feed looks a lot more vibrant and most of all having significantly less posts that show `0` comments really helps!

nexus/post.ts
```diff
@@ -103,14 +103,20 @@ schema.objectType({
     t.model.images()
     t.model.publishedAt()
     t.int('commentCount', {
-      resolve(parent, _args, ctx, _info) {
-        return ctx.db.comment.count({
+      resolve: async (parent, _args, ctx, _info) => {
+        const threadCommentCount = await ctx.db.comment.count({
```
Lanny, line 107: We can do these two count queries concurrently, ala 

```
const [threadCommentCount, postCommentCount] = await Promise.all([<queries here>])
```

Lanny (approved):
Think this is small change with good impact! One little implementation note, but otherwise looks good.

nexus/post.ts
robin-macpherson: Ah yes, great suggestion! 🙏🏼 🦔

PR #386, opened 2020-11-12
Enable users to update their handle

### LINKED ISSUE #373: Enable users to update their handle
### Perceived Problem

- Currently, users cannot update their handle.

- Let's get this working

#### Helpful Background/Context

- In the UI, this form can be found under settings > profile info
- We have an existing pattern and mutation in place for updating user information that you can follow
- See updateUser mutation and usage
- See the existing settings form for updating the user's name and email

Demezis:
## Description

**Issue:**: closes #373 

...

- I couldn't find a `bad request` response example so I made one (at least temporary),
- `SettingsForm` now shows more than one invalid input

## Subtasks

- [x] Edit updateUser mutation and enable handle change
- [x] Add form validation (frontend + backend)
- [x] I think I should clean up my code

## Screenshots
image


Lanny:
@Demezis See this is in a draft state. I checked it out on the preview env and the functionality looks good. If you have anything pending here feel free to leave it in draft and take as long as you need, just flip it over to ready for review when you'd like someone to take a look.

And as always thanks for the contribution!

robin-macpherson (approved):
Amazing work @Demezis! Thanks a lot for this 👏🏻 Will wait for @Lanny to take a quick look before merging

components/Dashboard/Settings/DetailsForm.tsx
```diff
@@ -106,11 +131,28 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ currentUser }) => {
...
                   name="handle"
-                  value={handle}
-                  className="j-field"
-                  onChange={(e) => setHandle(e.target.value)}
-                  ref={register({ required: t('profile.details.displayNameError') as string })}
-                  disabled={true}
+                  className={`j-field ${errors.handle ? 'is-invalid' : ''}`}
+                  defaultValue={currentUser.handle}
+                  ref={register({
+                    required: t('profile.details.displayNameError') as string,
```
Lanny, line 137: I'm guessing the return type of `t()` is `string | undefined` accounting for a missing translation, and thus the type assertion here. In this case I think something like `t('<string name>') || 'MISSING_STING'` or something that will show up noticeably in the UI would be preferable to the type assertion and possible throwing down the line.

nexus/utils/userValidation.ts
```diff
@@ -0,0 +1,73 @@
...
+    invalidFields.push({
+      name: 'handle',
+      message: 'profile.error.handleValidationErrorMessage',
+    })
+
+  return invalidFields
+}
+
+const CheckIfUnique = async (
+  db: any,
```
Lanny, line 26: I think `PrismaClient` is the type here, e.g. https://github.com/Journaly/journaly/blob/e98c199b6265a6f77760add2eba92c914686ba61/nexus/post.ts#L15

nexus/utils/userValidation.ts
```diff
@@ -0,0 +1,73 @@
...
+          NOT: {
+            id: userId,
+          },
+        },
+        or.length == 1 ? or[0] : { OR: or },
+      ],
+    },
+  })
+
+  if (email && users.some((u: any) => u.email === email))
```
Lanny, line 50: I like this approach to keep it to one query 👍

Lanny (commented):
Two minor things around types, but good stuff.

components/Dashboard/Settings/DetailsForm.tsx
Demezis: I changed this to the string interpolation. Just like it is in the `SignupForm`

Lanny (approved):

robin-macpherson:
🎉 🎈

Issue #370, opened 2020-11-06
Upgrade Nexus from v0.25.0 to v0.27.0 to enable Windows support


xferguson:
#### Description

It turns out I can't run `npm run build` on my machine because [Nexus@0.25.0 doesn't support Windows CLI](https://github.com/prisma-labs/graphql-framework-experiment/issues/965#issuecomment-645485119). They [released Windows support in July](https://github.com/prisma-labs/graphql-framework-experiment/pull/1184), and we could upgrade to the [Latest (0.27.0)](https://github.com/prisma-labs/graphql-framework-experiment/compare/0.25.0...0.27.0) which should work.

#### Repro Steps

1. Open a terminal on windows and run `npm run build`
2. **Expected**: The build should run and pass
3. **Actual**: The script will fail with an error: `The global Nexus CLI you invoked could not hand off to your project-local one becuase it wasn't on disk.`

xferguson:
I'm working on this first to ensure I can run that `npm run build`, and will open a PR soon (hopefully)

xferguson:
@Lanny @robin-macpherson I did the upgrade, but when I tested the `npm run build` I got this GraphQL error:

```
Error: nexus/post.ts:34:26 - error TS2322: Type '"PostTopic"' is not assignable to type '"ERROR: No subset types are available. Please make sure that one of your GraphQL type is a subset of your t.model('<ModelName>')" | undefined'.

34     t.model.postTopics({ type: 'PostTopic' })
```
Pretty weird, as I can see `PostTopic` defined 20 lines above it. Maybe it makes sense to one of you, but I will take a look at this again tomorrow when I have more time to code 😄

robin-macpherson:
@xferguson sorry for the delay getting back to you here, it's been a crazy couple of days! Hmm that error a bit strange, I haven't had a chance to look too closely yet but did you already check the release notes for the versions to see if there were any breaking changes? Since we were using the Nexus Framework at a very early stage, things were a little less stable between releases. That should be much better once we migrate off of it soon.

I can take a closer look tomorrow if you're still stuck 😊

Lanny:
[...]

PR #390, opened 2020-11-16
CircleCI Setup

Lanny:
## Description

**Issue:** Bandaid for: #370

Set up CircleCI to run the build commands as part of the commit-triggered build workflow

edit: looks like we're good to go, with a usable cache the build takes ~1.5 minutes, or 15 credits which means ~160 builds per week on the free plan. That should do for the time being.


robin-macpherson (approved):
Awesome job!! Aww, feels like our project is just growing up more and more 😢❤️

PR #391, opened 2020-11-18
Allow for nulls in email/handle of validateUpdateUserMutationData.

Lanny:
Fallout from recently merged PR. That PR was created before we had circle auto-builds set up so this should be something that's easy to catch before merging.


robin-macpherson (approved):
Awesome thanks for this! 💪🏼

PR #392, opened 2020-11-22
Add dialect to filter dropdown

### LINKED ISSUE #372: Show dialect for languages in Language Filter on My Feed Page
#### Problem

I just noticed that in the language filter dropdown in the my feed page, we don't show the dialect for languages that have one or more listed. For languages like Portuguese where we have both Brazilian & European, this results in seeing two entries for "Portuguese".

Screen Shot 2020-11-07 at 3 46 46 PM

#### Solution

Let's take the same approach that we have in the language form on the settings page where we show the dialect.
Since we already use parenthesis `()` to display the number of existing posts for each language, I would suggest using a hyphen instead for the dialect. So: `Portuguese - Brazilian (24 posts)`

This is how it looks on the settings page:

xferguson:
## Description

**Issue:** closes #372 

I added the dialect to the language dropdown for the feed as well in the "selected Languages" from the dropdown. It occurs to me that the postCard may need the dialect labeled as well. I wanted to check in about that before pushing that up (@robin-macpherson @Lanny).

I have not yet done any build (per the issues with nexus and windows), and if there are any problems with the circleci build, I will try and fix that with additional commits.

## Subtasks

- [x] I have added this PR to the Journaly Kanban project ✅ (already in the project)

### DIFF
--- a/components/Dashboard/MyFeed/MyFeed.tsx
+++ b/components/Dashboard/MyFeed/MyFeed.tsx
@@ -47,11 +47,14 @@ const MyFeed: React.FC<Props> = ({ currentUser }) => {
       hasPosts: true,
     },
   })
-  const languageOptions = (languagesData?.languages || []).map(({ id, name, postCount }) => ({
-    value: id,
-    displayName: `${name} (${postCount} post${(postCount || 0) === 1 ? '' : 's'})`,
-    selectedDisplayName: `${name}`,
-  }))
+  const languageOptions = (languagesData?.languages || []).map(({ dialect, id, name, postCount }) => {
+    const languageName = typeof dialect === 'string' && dialect.length > 0 ? `${name} - ${dialect}` : `${name}`
+    return {
+      value: id,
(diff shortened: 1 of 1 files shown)


Lanny (approved):
Spun up a vercel preview, looks good to me 👍

robin-macpherson (approved):
Awesome work, thanks @xferguson ! Regarding adding the dialect to the `PostCard`, that's a great thought but let's hold on that for the moment because I'm about to add the language level to the `PostCard` and I'd like to get that in first and we can see how much space we have to work with and what layout might work best. I'll have that done this coming week hopefully.

Excited to get this one in! 🎉

xferguson:
Thanks @robin-macpherson & @Lanny -- sorry I haven't had a lot of time to contribute. A lot going on at the moment, so I'm just grabbing "5-min issues" I can knock out when I have a sec. I'll try and carve out time for something a little bigger soon ☺️

PR #394, opened 2020-12-05
Nexus schema migration

Lanny:
## Description

Related to #370 , removes the nexus framework dependency, replaced with nexus-schema, and upgrades prisma and nexus-plugin-prisma.

## Migrations

No migrations


robin-macpherson (commented):
Epic stuff! Super excited to finally get past this big hurdle and also to finally have the ability to update our Prisma versions and take advantage of all the exciting developments there. I think for this one it would be super helpful for me to do a pair review just to go over the work and also strategy in terms of how we'll integrate this with the DB client work

robin-macpherson (approved):
WOOOOOOOOO OH BABY 🦔 🩰💨💥

PR #393, opened 2020-11-23
Wire up social links on settings page

Demezis:
## Description

**Issue:**: closes #367 

## Subtasks

- [x] I have added this PR to the Journaly Kanban project ✅
- [x] Add backend support for social media links
- [x] Social media links validation
- [x] Wire up social form
- [x] Wire up social media links on `ProfileCard`

## Migrations

- PR changes `SocialMediaPlatform` enum
-  `npm run prisma:generate` and `npm run migrate:up`

## Screenshots
image

PR #395, opened 2020-12-09
Use cross-env

Lanny:
## Description

So we can have the env vars set in scripts work on windows. With these changes I was able to run the build script and start the dev env on windows 10 (using git bash, but I don't _think_ that matters). Didn't get a chance to smoke test since I don't have postgres set up. Relates to #393 #370 . @xferguson if you could give these changes a try in your environment I'd be interested to confirm it works on windows.


robin-macpherson (approved):
Awesome, thanks for this @Lanny - let's hope this gets things working for @xferguson & @Demezis ! 🤞🏼

robin-macpherson:
Also @Lanny please see @Demezis 's recent comment on #393 regarding how he got things to work on Windows: https://github.com/Journaly/journaly/pull/393#issuecomment-742021835

xferguson:
Oops. I guess I'm a little late too the game here, @Lanny + @robin-macpherson 

1. Ran `npm ci` to update things.
2. Ran `npm run build` and go the error:
```
npm explore '@journaly/j-db-client' -- npm run prisma:generate
npm ERR! It doesn't look like '@journaly/j-db-client' is installed.
```
3. This seems to be unrelated to the original issue, so I hope that means the build will work for me normally. But it's weird that `j-db-client` isn't installed since I ran `npm ci` -- No related (seeming) errors in my install logs. Only these two:
````
Skipping 'fsevents' build as platform win32 is not supported
```
and
```
Error: There is a conflict between env var in ..\.env and .env
Conflicting env vars:
  DATABASE_URL

We suggest to move the contents of .env to ..\.env to consolidate your env vars.
```
^ That one is clearly from me doing something weird during the initial setup (it is unclear to me in the readme which folders should have a `.env`), but I'm not thinking it iis related to the build error.
4. Looked recent items. Found #374 + #381 -- @robin-macpherson, I see you worked on these. Am I going down the correct rabbit hole? There are no changes to the readme or the contribution doc, so I'm wondering if I need to just blow away my repo and start over? Or if there's a scipt or two I need to run that I'm missiing.
5. I did navigate to `/j-db-client` and run `npm ci` there, as I saw the `package.json`, but no change. Admittedly I was grasping at strings. I think I'm just missing some setup scripts.

robin-macpherson:
Hey @xferguson ! Yep so a day or two ago I merged in a pretty big change that pulls all of our Prisma stuff out into an internal DB client package so that we're able to install and use that in a variety of micro-services (such as `j-mail` so that w can do things like email batching for notifications). I was working on a little PR to update the docs accordingly and was going to tag you in the original PR to make you aware of the change but just didn't quite finish that yet. Sorry that this caused some confusion for you!

For now can you try simply moving your `.env` file to the root of the project and consolidating everything there, and can you try running `npm i` just out of curiosity?

PR #398, opened 2020-12-15
Enable feed search

### LINKED ISSUE #107: Implement My Feed Search, Filtering, and Pagination

Demezis:
## Description

**Issue:** closes #107 

Hi. It's me again 😄
I have tried to split feed into smaller elements, but I don't know yet how to do this properly with functional components to prevent re-render. 🤔 
Let me know what you think.

## Subtasks

- [x] Feed search by phrase and topic
- [x] Post index

## Migrations

**migrate:up**

- Does the migration need to be applied before or after deploying code? **It doesn't matter**
- Is applying the migration going to necessitate downtime? **No**
- Is there any SQL or backfill logic that has to be run manually? **No**

## Screenshots
journaly


robin-macpherson:
Awesome, thank you @Demezis ! I started reviewing and will try to finish that and post any comments tomorrow! (Thursday) 😄

components/Dashboard/MyFeed/MyFeed.tsx
```diff
@@ -32,37 +32,20 @@ type Props = {
...
-   */
+  const [search, setSearchState] = useState('')
+  const [selectedTopicsFilters, setSelectedTopicsFilters] = useState<number[]>([])
 
   // Fetch languages that have at least 1 post
   const { data: languagesData } = useLanguagesQuery({
     variables: {
       hasPosts: true,
     },
   })
```
robin-macpherson, line 43: We're going to need to update the `topics` query in `topics.ts` so that it uses the `hasPosts` argument to filter only topics that have existing posts and can then be used in the same way that we fetch only languages that have at least one post here in `useLanguagesQuery`. Currently we end up with just all topics in the dropdown including ones that have no posts

j-db-client/prisma/schema.prisma
```diff
@@ -120,6 +120,8 @@ model Post {
   threads      Thread[]
   postTopics   PostTopic[]
   postComments PostComment[]
+
+  @@index(fields: [title, body], name: "title_body_index")
```
robin-macpherson, line 124: Great effort here but after having a brief look I don't think this is quite going to work in the way that we want. We'll have more concrete suggestions on this soon after taking a closer look

robin-macpherson:
Great stuff @Demezis ! I haven't had a chance to leave a full review yet but just left one important comment around filtering topics that will need to be addressed so that you have some feedback for now.

There's quite a lot here and our general approach has been that we try to contain a piece of work to a PR, so I think my approach would have been to have separate PRs for (1) the work on splitting up/reorganizing the `MyFeed` page, (2) getting the search bar working, (3) Getting the topics filter working, and possibly even a separate little PR for the indexing, just in case something goes wrong and we want to easily roll it back without having to revert all the other work as well. This makes it easier to digest, review, and get the work into production more quickly.

All the work looks excellent though and I'm **very** excited to get everything on the MyFeed page working! 💪🏼😄 Just wanted to share some thoughts on our general approach.

Thanks again for the amazing work and I'll be back to finish the review or add more comments as soon as I an

Demezis:
I understand. I will keep that in mind.

j-db-client/prisma/schema.prisma
Demezis: I think I see now, there are two index types for text search in PostgreSQL: GiST and GIN, and... prisma doesn't support them.

robin-macpherson: Yeah it seems so. I'll talk with the folks at Prisma and see if maybe we can help to get support into a coming release

robin-macpherson:
It's really exciting to see the My Feed page working with search and topic filters @Demezis ! Great work 😄

While testing on `staging` I did just notice that the "my languages" filter is broken, it seems to always be on and there is no way to clear it (in terms of the UI) but it also doesn't apply any languages to the filter

robin-macpherson:
Hey @Demezis , thanks for pushing up the fix for the "my languages" filter and also for updating topic query.

I just looked through things again and there's just one last thing I'm seeing now which is that as things are currently, the `topics` query isn't aware of the languages filter. So if you select "Hebrew", for example, the topics filter still returns all topics with their post count so the user can then go and select a topic that has no posts available in that user's language

robin-macpherson:
Hi @Demezis , is this ready for re-review? I noticed you have been pushing commits after my comments your last comment on the PR was 5 days ago so I haven't been sure exactly what the status of the PR is. I just tested this again because I saw the new commit but it looks like it might not be complete yet, I'm not sure.

It would be super helpful if you can let me know whenever you are ready for re-review. Thank you! 😄

Demezis:
@robin-macpherson Yes, please re-review it. I hope it is fine now 😄

robin-macpherson:
Great thank you @Demezis ! I think the only remaining thing now is that you are still returning the full list of topics even if there are 0 posts. The ideal behavior would be that we only show topics in the filter dropdown if they have at least 1 post.

Does that make sense?

So basically all of the filters should be connected to each other and updating one filter should update each query

Demezis:
So when I change topics, the languages list should be filtered too? Am I understanding it correctly?

robin-macpherson:
> So when I change topics, the languages list should be filtered too? Am I understanding it correctly? Or languages with 0 posts should remain?

That's what I am thinking. Do you have any thoughts on this behavior @Lanny ?

Demezis:
I guess that as a user, I want to see all possible topics/languages options, but the options with 0 posts should be marked as disabled. Does that make sense?

robin-macpherson:
> I guess that as a user, I want to see all possible topics/languages options, but the options with 0 posts should be marked as disabled. Does that make sense?

Yes I can see your point, this would help. I guess for me the problem is that we have a _**lot**_ of topics and that list is going to get longer and longer. If a language only has 3 available topics, I feel like it might be frustrating for the user to have to scroll through 150 topics just to check if there are any with > 1 post . What do you think? I'm definitely open to discussing it, this is a tricky one

Demezis:
Disabled options would be on the bottom of the selected list

Demezis:
I uploaded a version with disabled options, please take a look. Let me know if you will have some ideas.

robin-macpherson (approved):
> I uploaded a version with disabled options, please take a look. Let me know if you will have some ideas.

That’s a great point regarding disabled options going at the bottom. I think your solution works great for now! 😄 I think I will want to iterate on this in the future but for now the most important thing is getting everything fully functional and I think this is great. This way we can also survey some users to get their thoughts, too.

Thanks again for the great work on this @Demezis and sorry that I was a bit slow with the reviews. It’s been really crazy managing everything but things are getting better! 😄

I will do a little more testing and then merge this in 🚀

robin-macpherson:
@Demezis I just did one last round of testing and it looks like the results with 0 posts are not at the bottom. The current behavior is what I was worried about, I'll attach screenshots:

Screen Shot 2020-12-22 at 1 22 13 PM


image

Demezis:
@robin-macpherson I see... 
image

I don't have any other ideas right now.

robin-macpherson:
@Demezis ohhhh ok sorry, must have had some weird caching or something. I just cleared everything out and now I see that behavior. Looks great! 🙌🏼 Merged 🙂

PR #399, opened 2020-12-17
Build PostComment Subscriptions

### LINKED ISSUE #382: Subscriptions for Post Comments
#### Perceived Problem

- Currently, we don't have a Subscriptions implemented for Post-level comments
- This means that users aren't notified when somebody responds to their post comments
- We're finding that these post comments are actually quite popular

- Implement subscriptions for post comments
- backfill as we did with Thread Subscriptions

robin-macpherson:
## Description

**Issue:** closes #382 

- Builds subscriptions for `PostComment`s so that anyone who leaves a `PostComment` is subscribed and is notified of other post comments from that time on.
- Post authors are subscribed at the moment of creating a new post.
- For SQL to perform data migration, see the **Data Migration** section below which has also been stored in in the migration README.

## Subtasks

- [x] Update schema
- [x] Update post creation to subscribe author
- [x] Update `PostComment` creation to subscribe comment author & notify all subscribers
- [x] Custom SQL

## Migrations

_This PR contains a migration **AND** requires a data migration to backfill the `PostCommentSubscription` table with all existing post comments. Data migration SQL is below._

  - **YES**
  - **While the backfill is running**
  - **YES**
  - **I would say it is worth taking a manual DB snapshot since we are going to be backfilling data for all `PostComments` in our DB.** Would like to get your thoughts @Lanny

## Data Migration

```sql
-- Subscribe participants in PostComments of a Post to that post for future comments
INSERT INTO "PostCommentSubscription" ("userId", "postId", "createdAt") (
  SELECT "authorId" AS "userId", "postId", NOW() AS "createdAt"
  FROM "PostComment"
  GROUP BY "userId", "postId"
) ON CONFLICT DO NOTHING;

-- Subscribe post authors to all PostComments on their Posts
INSERT INTO "PostCommentSubscription" ("userId", "postId", "createdAt") (
  SELECT "Post"."authorId", "PostComment"."id", NOW()
  FROM "PostComment"
  JOIN "Post" ON "PostComment"."postId" = "Post"."id"
) ON CONFLICT DO NOTHING;
```
```

### CODE REVIEW

File: j-db-client/prisma/schema.prisma [hunk 1]
```diff
@@ -8,39 +8,40 @@ generator prisma_client {
 }
```
robin-macpherson, line 8: @Lanny - Prisma seems to prefer declaring `@id` first so any one-line changes simply switching that are just to satisfy the Prisma schema formatter plugin. Same for whitespace changes.

j-db-client/prisma/schema.prisma
```diff
@@ -8,39 +8,40 @@ generator prisma_client {
...
+  threadSubscriptions      ThreadSubscription[]
+  postLikes                PostLike[]
+  commentThanks            CommentThanks[]
+  postCommentThanks        PostCommentThanks[]
+  comments                 Comment[]
+  postComments             PostComment[]
+  badges                   UserBadge[]
+  followedBy               User[]                    @relation("UserFollows", references: [id])
+  following                User[]                    @relation("UserFollows", references: [id])
+  postCommentSubscriptions PostCommentSubscription[]
```
robin-macpherson, line 40: New field added

j-db-client/prisma/schema.prisma
```diff
@@ -89,59 +91,61 @@ model LanguageNative {
...
+  readTime                 Int                       @default(1)
+  likes                    PostLike[]
+  language                 Language                  @relation(fields: [languageId], references: [id])
+  languageId               Int
+  longitude                Float?
+  latitude                 Float?
+  threads                  Thread[]
+  postTopics               PostTopic[]
+  postComments             PostComment[]
+  postCommentSubscriptions PostCommentSubscription[]
```
robin-macpherson, line 125: New `postCommentSubscriptions` field added

j-db-client/prisma/schema.prisma
```diff
@@ -256,24 +263,36 @@ model Image {
...
+  id        Int      @id @default(autoincrement())
+  createdAt DateTime @default(now())
+  user      User     @relation(fields: [userId], references: [id])
+  userId    Int
+  post      Post     @relation(fields: [postId], references: [id])
+  postId    Int
+
+  @@unique([userId, postId])
+}
+
```
robin-macpherson, line 293: The Gorgonzola of the schema work 🧀 🐁

nexus/post.ts
```diff
@@ -348,6 +348,23 @@ const PostMutations = extendType({
           },
         })
 
+        // automatically subscribe the post author to all PostComments made on their post
```
Lanny, line 351: I think we can do this as a nested write when we create the post and save a round trip to the database

Lanny (approved):
One ask but looks great!

nexus/post.ts
robin-macpherson: Done!

PR #402, opened 2020-12-23
Enable users to update their location

### LINKED ISSUE #330: Enable "Location" to be updated
#### Perceived Problem

- Currently, users can't update location because it's tied to our `Location` model which is great and is setup for proper locations, but building out the frontend mechanism to capture that information isn't a high priority
- We believe there is still value in simply being able to see someone's location

- Break "location" field into two inputs: "City" & "Country" on the frontend
- Remove `Location` model and simply add two fields to `User` for city & country

### LINKED ISSUE #403: 🐞 Details / Settings Form Validation BUG: Handle
#### Description

There is a bug with the settings details form where when you try to submit the form you get a validation error that the `handle` must be at least 3 characters even when you have a handle that is clearly more than 3 characters long

image

#### Repro Steps

1. For me I literally just want to my settings and attempted to save the form without making any changes

robin-macpherson:
## Description

**Issues:**
- closes #330 
- closes #403 

Currently users are unable to update their location. The original schema design was more robust with how we planned to store location data but setting this up properly would require quite a lot of work relative to the priority of simply allowing users to indicate their location vs. other work we have planned. The short/medium term solution here is to simplify things by removing the `Location` model and simply adding two new fields to the `User` model: `city` and `country`.

## Subtasks

- [x] Update schema
- [x] Generate migration
- [x] Update nexus stuff
- [x] Update settings form and get things working
- [x] Externalize strings for i18n
- [x] Update profile to use the location

## Migrations

- Does the migration need to be applied before or after deploying code? **BEFORE**
- Is applying the migration going to necessitate downtime? **NEGATIVE**
- Is there any SQL or backfill logic that has to be run manually? **NEGATIVO**
- Are the DB changes high risk in your estimation? Should a manual DB snapshot be taken before applying? **LOW-RISK**

## Screenshots

update-location


components/Dashboard/Settings/DetailsForm.tsx
```diff
@@ -203,16 +207,30 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ currentUser }) => {
               </div>
               <div className="details-form-field">
                 <label className="settings-label" htmlFor="location">
```
Lanny, line 209:
```suggestion
                <label className="settings-label" htmlFor="city">
```

components/Dashboard/Settings/DetailsForm.tsx
```diff
@@ -203,16 +207,30 @@ const DetailsForm: React.FC<DetailsFormProps> = ({ currentUser }) => {
...
+                  name="city"
                   className="j-field"
-                  placeholder="Coming soon..."
+                  placeholder={t('profile.details.cityPlaceholder')}
+                  defaultValue={currentUser.city || ''}
+                  ref={register()}
+                  />
+              </div>
+              <div className="details-form-field">
+                <label className="settings-label" htmlFor="location">
```
Lanny, line 223:
```suggestion
                <label className="settings-label" htmlFor="country">
```

Lanny (approved):

PR #405, opened 2020-12-28
Post LanguageLevel Improvements / DB & UI

### LINKED ISSUE #404: Display Language Level On Posts/PostCards
#### Perceived Problem

- A lot of users have reported they would get great value from being able to see what level a post was written at
- When implementing language levels, we identified that in order to display language levels on posts, the best option would be to record what a user's level was at the time of writing it.

- I propose recording the user's language level at the time of **publishing**, similarly to how we record `publishedAt`
- Add corresponding field to `Post` model
- Add the `publishedLanguageLevel` to `PostCard`s and in the header of posts

### LINKED ISSUE #406: 🐞 Post Deletion Broken: Cascading Delete issue with introduction of PostCommentSubscriptions
#### Description

Currently, deleting a post fails in production after the introduction of `PostCommentSubscription` because I forgot that we have to manually handle our own cascading deletes.

Fixing in #405

robin-macpherson:
## Description

**Issues:** 
- closes #404 
- fixes #406 

Implements mechanism for capturing `publishedLanguageLevel` as well as displaying that language level in `PostCard`s on the My Feed page.

Includes migration to add that field to the `Post` model.

**NOTE:** I originally was going to add the language level to the `PostHeader` as I know we discussed that, but after implementing it I felt quite conflicted about the design as the header felt quite overcrowded with badges. For now I'm thinking we just start by putting it in the post card but am open to other ideas.

**NOTE 2:** I also found a production bug while testing this one and am including the fix here for expediency (see #406 )

## Subtasks

- [x] Update schema
- [x] Implement BE logic
- [x] Update Apollo GQL hooks
- [x] Implement UI

## Migrations

- **_Migration will need to be applied before merging the code_**
- No downtime required
- _**This will require a manual SQL backfill to update all `Post` records so that the `publishedLanguageLevel` column matches up with the users corresponding `LanguageRelation` records**_
- There is some risk here if we screw up the backfill SQL, but also in case some users have not updated their language levels all of their posts will remain marked as "BEGINNER"

## Screenshots
image


robin-macpherson:
@Lanny posting for early feedback on the BE approach

PR #407, opened 2020-12-29
Fix 404 on thread click

Lanny:
Was never able to repro in local but looking at the stack trace I got the one time I could repro on prod I think this should fix it. The `popover-root` portal container appears to be null in some situations, so we need to handle that case instead of waving away the null case with a type assertion.

## Migrations

No migs


robin-macpherson (approved):
Woohoooooo thanks for this! I literally just came to create the issue for ya and saw you already had the PR up haha. Hedgehog gives you a 10/10 🎈🦔

PR #414, opened 2021-01-01
Fix issue where emails would always be sent to author of a post comment when a new one is made

Lanny:
## Description

See title, we were just passing the wrong user into "sendPostCommentNotification", seemingly a holdover from the previous "only notify the post author" implementation. Also slight copy change to that email to make it correct for non-post-authors

## Migrations

No migs.

### DIFF
--- a/nexus/comment.ts
+++ b/nexus/comment.ts
@@ -366,7 +366,7 @@ const PostMutations = extendType({
 
           const promise = sendPostCommentNotification({
             post,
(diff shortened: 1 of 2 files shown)


robin-macpherson (approved):
Agh, bloody hell 😓 What a mistake lol bravo me 👏🏻

Thanks so much for the fix! 🥂 🦔

PR #415, opened 2021-01-02
Stop handing out beta badge.

Lanny:
## Description

Not in the closed beta period anymore, so we don't need to create users with this badge


robin-macpherson (approved):
Good call!

PR #420, opened 2021-01-04
Move #popover-root to _document.tsx

Lanny:
Relates to #407 , core issue is `#popover-root` is not present when a thread is opened and it needs to be. It's unclear to me why that element doesn't get rendered sometimes, but it seems to also affect `<ToastContainer>` (also in `_app.tsx`) but not `#modal-root` (in `_document.tsx`) so let's just try moving lifting it up out of `_app.tsx` and see what happens

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -29,7 +29,6 @@ class JournalyApp extends App {
         </Head>
         <Component {...pageProps} />
         <ToastContainer />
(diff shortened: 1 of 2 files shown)


robin-macpherson (approved):
Interesting! I just clicked around a bunch of posts on staging and didn't encounter the issue at all, fingers crossed! 🤞🏼

PR #422, opened 2021-01-05
Make Sign Up Errors More User Friendly

### LINKED ISSUE #421: Improve "Unique Constraint" Error Message In Sign Up Form
#### Description

- We've had a few users who seem to have signed up and then continued trying to sign up, perhaps through clicking the sign up button twice?
- The error message the user sees is the "unique constraint" backend/DB error directly.
- Let's return a more user-friendly message, such as "This email address already exists. Please try logging in"

image

robin-macpherson:
## Description

**Issue:** closes #421 

A surprising number of users have been confused into thinking there was a major bug with our app because they were either accidentally clicking sign up twice or using a handle that was unavailable, because we were showing them the unique constraint DB error.

This PR adds a couple of checks to handle those errors better.

## Subtasks

- [x] Handle `createUser` mutation

### DIFF
--- a/nexus/user.ts
+++ b/nexus/user.ts
@@ -148,10 +148,24 @@ const UserMutations = extendType({
         email: stringArg({ required: true }),
         password: stringArg({ required: true }),
       },
-      resolve: async (_parent, args, ctx: any) => {
+      resolve: async (_parent, args, ctx) => {
         if (!args.handle.match(/^[a-zA-Z0-9_-]+$/)) {
           throw new Error('Invalid handle')
         }
+
+        let existingUser = await ctx.db.user.findUnique({
+          where: {
+            handle: args.handle,
+          },
+        })
+        if (existingUser) throw new Error("Handle is already in use")
+
+        existingUser = await ctx.db.user.findUnique({
+          where: {
+            email: args.email,
+          },
+        })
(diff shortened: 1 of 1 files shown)


nexus/user.ts
```diff
@@ -148,11 +148,25 @@ const UserMutations = extendType({
...
+          },
+        })
+        if (existingUser) throw new Error("Handle is already in use")
+
+        existingUser = await ctx.db.user.findUnique({
+          where: {
+            email: args.email,
+          },
+        })
+        if (existingUser) throw new Error("Email address is already in use. Please try logging in")
```
Lanny, line 168: I'd use `ResolverError` or one of its subclasses from `nexus/errors` here. Maybe `UserInputError` which I think is what we use when changing handles.

Lanny (changes requested):
Trying to pre-check that the `user.create()` call won't violate a unique constraint creates a race condition. Usually in this situation you just want to try to do the insert and handle failure, avoiding a race. I think what you want to do is something like:

```
try {
  const user = await ctx.db.user.create({...})
} catch (ex) {
  if (ex.code === 'P2002') { // Prisma's error code for unique constraint violation
    throw new UserInputError('Email or handle is already in use, please select another')
  } else {
    throw ex // Rethrow for debugging purposes
  }
}
```

Lanny (approved):
👍

Issue #400, opened 2020-12-22
Reorg repo to have more consistent directory structure


Lanny:
Right now the root level of the repo has a big ole mix of conceptual levels of code, some example of inconsistencies:

- `nexus` contains pretty much all the "backend" logic
- The "frontend" code is spread across `lib`, `elements`, `generated`, `graphql`, `components`, `config`, `hooks`, `theme`, `utils`, and `pages`
- `j-db-client` and `j-mail` are entirely separate deployments
- `docs` and `cypress`, while sparse, should probably be above everything else

Here's a proposed organization, totally open for discussion:

- ROOT
  - packages
    - j-mail
    - j-db-client
    - web
      - components
        - (all the existing stuff in components)
        - (all the stuff in elements, the components/elements separation never made a lot of sense to me)
      - pages
      - nexus
      - utils
        - (all the existing stuff in the top level utils)
        - i18n.js (moved from config, kill that dir)
      - generated (should probably get this out of source control at some point)
      - hooks
      - lib
      - public
      - scripts(or at top level?)
      - styles
      - theme
      - tsconfig.json
      - vercel.json
      - next*
      - codegen.yml
  - docs
  - cypress
  - README.md
 
We'll need to see if we can get vercel to deploy with the vercel stuff not being at the root dir, and some of the circle jobs will need to be re-written.

We could also consider a harder separation of the FE and BE (we ended up merging the two for deploy reasons, we could revisit and see if what we've learned in the last several months makes this surmountable) but I don't think it's a big deal either way.

robin-macpherson:
We've talked about this in person but for visibility just a quick comment to say I think this makes a lot of sense and especially for new contributors I think this will make digesting the codebase and structure of things far more intuitive. I'm also fine with just consolidating `elements` and `components`

robin-macpherson:
One more note, we've confirmed that Vercel has now added the option for setting the base directory for running deployment scripts, etc. so that removes the main blocker we had before while we're still on Vercel for now.

PR #436, opened 2021-01-09
Approach Doc - Containers

JoshuaBelden:
## Description

This is a request to discuss a new containerized approach to using postgres in a local development environment. I realize that the team is heads down on more important matters and don't have an expectation that this will get attention anytime soon. 

This is more about letting you all know, I did a thing with docker and here's what it looks like. I'm extremely new to this project and probably have a lot of naive thoughts about the environment, so if this was already discussed and turned down, please disregard and I can close this out.

**Issue:** currently no issue

...

https://www.journaly.com/dashboard/profile/3068


docs/approach-docs/README.md
```diff
@@ -21,3 +21,4 @@ The existing types are:
 
 - `system-` for a `system` or set of conventions
 - `feature-` for creating a new or enhancing an existing `feature`
+- `platform-` for infrastructure or process level improvements
```
JoshuaBelden, line 24: Not sure this is the most appropriate name, `infrastructure` was my other opinion

robin-macpherson:
Hi @JoshuaBelden , thanks for the approach doc and for looking into this!

Journaly used to actually be fully containerized with Docker and at the time I was also experimenting with Swarm for orchestration, but we ended up going away from that approach. I'm happy for us to discuss it further if you would like, and both @Lanny and I are totally fine with having this alternative/optional approach available to developers for setting up their local database.

I'll take some more time soon to look over the details of your set up. In the meantime, I'm not sure if this is helpful or interesting but here are a couple of PRs from way back when we first introduced Docker and when we changed our approach. The codebase has changed enormously since then.

It's nostalgic looking back at them!

https://github.com/Journaly/journaly/pull/9
https://github.com/Journaly/journaly/pull/91

docs/approach-docs/platform-containers.md
```diff
@@ -0,0 +1,98 @@
...
+
+You'll see that I have a bootstrap shell script to set up some initial database work:
+
+```sh
+#!/bin/bash
+
+set -e # bail immediately on error
+
+psql <<- EOSQL
+    CREATE USER developer WITH CREATEDB PASSWORD 'nights_watch';
```
robin-macpherson, line 33: haha you've already assimilated into our culture very well! 🤓

Lanny:
I never realized we had approach documents! That's cool, although I guess we haven't used them too heavily before now

My two cents on dockerization is that I totally see the developer benefit, DB setup is the biggest section of the setup process in the readme and we've had some headaches trying to support running on windows natively (I don't even know that prisma query engine has windows binaries).

WRT making a journaly instance runnable in prod via docker, that's harder. We've always assumed deployment to a serverless style environment  rather than a long running traditional "box server" kind of situation. For the web code you can change some next build params (and probably several other things about deployment) and get it going, but stuff like j-mail would require a pretty significant refactor to be able to dockerize.

tl;dr 👍  from me for improving the developer experience, but don't sweat trying to get production ready deployments going with docker unless you _really_ want to

docs/approach-docs/README.md
robin-macpherson: Good point! I think `infrastructure` probably fits best with the terminology we often use but platform works fine as well!

docs/approach-docs/platform-containers.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+You'll see that I have a bootstrap shell script to set up some initial database work:
+
+```sh
+#!/bin/bash
+
+set -e # bail immediately on error
+
+psql <<- EOSQL
```
Lanny, line 32: I like the approach of seeding the DB on build. `npm run db:snapshot` can be used to dump a development DB, we could check in the `snapshot.sql` file that generates and use that to seed the DB

robin-macpherson:
Thanks for the input @Lanny ! I totally agree in supporting anything to improve the local development set up, and standing up a local DB instance in particular is incredibly simple and easy for people once that's al set up. I was about to look into your issue with data persistence @JoshuaBelden but unfortunately am getting pulled into something urgent at work. Perhaps you can look at how we had our volumes set up in our old Dockerfiles for some clues? It seems like you've got most of this figured out really well!

JoshuaBelden:
@robin-macpherson I'll definitely take a look at the previous pull requests. While it's easy to imagine how to dockerize the database, I was wondering about other aspects of the app, these PRs should provide some good context.

I agree @Lanny dockerizing everything in production is probably not the right direction to go in.

docs/approach-docs/platform-containers.md
```diff
@@ -0,0 +1,98 @@
...
+
+```
+
+### Final Modifications
+
+With the given files, we end up with the following additions:
+
+- `db-bootstrap.sh`
+- `Dockerfile`
+- `docker-compose.yaml`
```
Lanny, line 71: The proposal will dockerize the DB but not the actual app process, e.g. `npm run dev`. I'd like to add another container that will dockerize the actual development server. The benefits there would be:

- `docker-compose up` will bring up both the DB and the dev server
-  we won't have to fuss with cross platform support for prisma and nexus. Also npm scrips use cross-env and we could probably get away with kicking that dep if we could say "dockerized env is the supported dev setup on windows"
- we could run some of the one off stuff like `prisma generate` and nexus reflection as either part of the build or startup phase, so just that many less hoops for developers to jump though

We could do the work in two passes, just the DB then the later the application itself, but I think there are some some definite wins in doing both

docs/approach-docs/platform-containers.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Docker Knowledge
+
+I found docker difficult to grasp when it came out and still struggle with advanced concepts. We'd be asking our developers to have some level of understanding of docker to work with our stack. If they don't have docker at all, it would require an install of additional software.
+
+At this point, it would be optional, and an alternative to installing postgres.
+
+### Permissions
+
+You'll notice that the `docker-compose.yml` file has a line for the postgres setting `POSTGRES_PASSWORD`. This is required for postgres to run, which means yet another password for the developer to setup. The alternative is to set postgres to run completely insecure, not a great option even though it's for local development.
```
Lanny, line 92: I'm OK with this being insecure since we probably won't be productionalizing the docker env. We can just hardcode a password. If we dockerize the app as well, I don't think we need to `EXPOSE` postgress outside of the docker compose sphere so it won't really matter if the password is common knowledge, no one outside of the dockerized env will be able to connect to PG.

docs/approach-docs/platform-containers.md
```diff
@@ -0,0 +1,98 @@
...
+
+Also, the npm script that runs for `reseed` deletes tables, and for some reason, it doesn't have permissions with this setup. I would need to research the problem further, however, now that we're in a container, we could simply re-build the image, eliminating the raw SQL in the package.json file.
```
Lanny, line 94: oh yeah, `npm run empty-db` (which is run as part of `npm run reseed-db`) sources your `.env` file, so if the `DATABASE_URL` var there isn't updated to point at the dockerized instance (or if it does point to the right place but doesn't have valid credentials) it would probably fail to do its thing

docs/approach-docs/platform-containers.md
robin-macpherson: Great idea!

docs/approach-docs/platform-containers.md
robin-macpherson: I agree with this! In this case, I am wondering if I should try to tackle the project re-org first then? I was going to try and do it in the next day or two and see how seamless it ends up being? #400

Lanny: Yeah, re-orging first would probably make our lives simpler. "web" (or whatever we call the main thing) having its own dir instead of just being the project root would make sense as a place for a docker file, and the compose file could live at the actual root.

robin-macpherson:
@JoshuaBelden ok so I'll work on the project re-org that @Lanny proposed in #400 first in order to make our lives easier here, and then we can work on getting everything working here just so that we don't create extra work and it should make things simpler. I'll do my best to get to that in the next day or so and hopefully we don't run into any snags. Will loop you into the PR as well

docs/approach-docs/platform-containers.md
JoshuaBelden: The only issue I see is if a developer wants to query or update local dev data with a client, they won't be able to connect to it. Also, npm scripts for database updates wouldn't be able to connect.

JoshuaBelden:
Sounds good. I've got an additional small update to rename `project` to `infrastructure`, otherwise I'll put this down for now until we're over the re-org.

docs/approach-docs/platform-containers.md
Lanny: The workflow I'm used to is that you `docker exec` into a container to run stuff like NPM scripts or a DB shell, although I suppose it is kind of an annoying extra step.

I'm good either way, having to exec into a container to run some dev tools doesn't really bother me, but I also don't think opening up a DB running in a container to whatever network environment your development setup is in (in most cases, probably your home router) is a huge security concern either.

robin-macpherson:
Hi @JoshuaBelden ! Very sorry for the delay, we had quite a few really high-priority items come up but I just wanted to update you that I am working on the project reorg now, and should have things ready for you to work on this in the next couple of days. Thanks so much for being patient with us!

JoshuaBelden:
@robin-macpherson Thanks for the update and no worries about a delay. I'm going to be leaving for vacation in a couple days but will be back next week and can touch base and see what's next.

robin-macpherson:
> @robin-macpherson Thanks for the update and no worries about a delay. I'm going to be leaving for vacation in a couple days but will be back next week and can touch base and see what's next.

Sounds good @JoshuaBelden ! Just FYI this PR will be the one that completes that work. Still a work in progress but hoping to finish it today. Have a lovely vacation! 🎉

robin-macpherson:
Hi @JoshuaBelden ! Just wanted to check in and see if you would still like to work on this? I hope you had a great vacation!

JoshuaBelden:
Hi @robin-macpherson , good to hear from you again. I think I might be taking too much vacation time. :) I'm headed out into nature tomorrow, but will be back on Monday. I'll see if I can get this branch updated and look into moving it forward.

robin-macpherson:
Haha sounds great, @JoshuaBelden ! No rush, I just wanted to check in on it. Have a wonderful time out in nature :)

JoshuaBelden:
@robin-macpherson So I've spent some time getting my head back into this. I have a local branch I've been working on the containerization on and I have that all up to date with your latest changes in the master branch. 

As it stands now, I only have the database in a container with a bootstrap that sets up the journaly user. All other setup remains the same, including using prism to setup the database once the image is running. I saw a few suggestions for improvements but not sure which ones we should put in place.

As this is just an approach doc, should I be committing actual code changes here, or do we want to start dev on an actual issue?

JoshuaBelden:
@Lanny That sounds good. There was mention of containerizing the service but wasn't sure if we wanted to hold on that or not. I think it makes sense to do both. I'll do a bit of local work to see what that looks like and then update the approach doc so we can start work.

I may need to lean on you for actual implementation if anything veers from docker happy path. :) I'm still coming up to speed.

JoshuaBelden:
@Lanny @robin-macpherson I've updated the approach doc to include a container for the node server. I'm using docker-compose to setup postgres and the docker file is now focused on running node. 

There were a couple things I started stumbling on:

- Multiple docker files not-supported. I believe we can use docker-compose to orchestrate the postgres setup.
- Bootstrapping the database from docker-compose remains a mystery that I can figure out in development.
- The root readme.md appears to be off a bit, it makes mention of the path `journaly/packages/web` but that folder is missing. 

If this all looks OK to move forward, I'll wait to see an issue created and assigned to me and I can start on it. Also, do you have a slack channel for general discussions, or do we use GitHub for communication?

Lanny:
>Multiple docker files not-supported. I believe we can use docker-compose to orchestrate the postgres setup.

Oh yeah, I kind of remember this being funky. Last time I used docker-compose the solution was to structure the directory like:

```
|- Docker
| |- web
| | |- Dockerfile
| |- db
|    |- Dockerfile
|- ...everything else
```

then in the compose file something like:

```yaml
services:
  web:
    build: Docker/web
  db:
    build: Docker/db
```

Alternatively if you want all the dockerfiles at the top level you can do something like:

```yaml
services:
  web:
    build:
      context: .
      dockerfile: Dockerfile-web
  db:
    build:
      context: .
      dockerfile: Dockerfile-db
```

(this is based on [here](https://docs.docker.com/compose/compose-file/compose-file-v3/#build) but I haven't actually tested anything)

> The root readme.md appears to be off a bit, it makes mention of the path journaly/packages/web but that folder is missing

I think the assumption there is that the root dir of the repo is named "journaly" (default when cloning)

This looks good to me, I'm approving but will wait for @robin-macpherson to merge.

> Also, do you have a slack channel for general discussions, or do we use GitHub for communication?

We do have a slack but it's only the core team at this point. I would love to open it up to anyone who wanted to join, but I think there are some questions about how that interacts with pricing and if we can limit channel access (we've assumed that it's private up to now and there are probably private keys and stuff hanging out in some of the channel histories). @robin-macpherson I'm led to beleive there's a way to create an open invite link from [this page](https://journaly.slack.com/admin/shared_invites) but I'm not a slack admin so I can't verify. Whenever you get a chance could you take a peek and see how that works with pricing and channel access?

Lanny (approved):

robin-macpherson:
Awesome work with the approach doc, @JoshuaBelden ! 
@Lanny yes we can have what Slack calls "guests" but multi-channel guests get billed as normal members, so I would have to pay for every new person who joined and that would become really expensive. You can have single-channel guests for free, but only if you're a paying organisation. We're currently on a free trial and I'm still figuring out whether it's feasible for me financially to be paying for Slack in addition to our other stuff until we start generating a little bit of revenue, so I think let's wait until we see what happens there and once we commit to paying for Slack, we could create a single guest channel and invite contributors to join it as part of our little welcome message. What do you think?

robin-macpherson (approved):

JoshuaBelden:
I appreciate the Slack considerations but please don't incur additional charges just to get me into slack. I'm happy to have conversations inline here.

robin-macpherson:
> The root readme.md appears to be off a bit, it makes mention of the path journaly/packages/web but that folder is missing.

Just to be sure, I assume you've updated your fork?  If not you may not have the changes from the big project re-org (#497) that we discussed a bit earlier up the thread of this PR. But if so then it may be what @Lanny mentioned in his last comment. For me the route is `journaly` (default when cloning as Lanny mentioned) so the path is `journaly/packages/web`

robin-macpherson:
@JoshuaBelden I've created an issue for you, would you mind just leaving any comment on #568 so that I can assign it to you? I'm unable to assign directly to users outside the core team unless you've commented on the issue.

PR #438, opened 2021-01-11
Some last minute changes to j-mail

Lanny:
Changes update cron schedule to be once daily (it was already deployed like this, but I never committed the change), and fixes the issue where gmail shows the update email as a set of columns.


robin-macpherson (approved):
### **THANK YOU!** 🌾... 🦔💨💥

PR #439, opened 2021-01-11
Only include published posts in `postsWrittenCount`.

Lanny:
Super small one, just what it says on the tin.


robin-macpherson (approved):
Great catch!

Issue #416, opened 2021-01-03
When editing an entry, the text presented is an older version than the one in display mode


rickyweb:
|                    |                      |
| ------------------ | -------------------- |
| repro              | [link](https://todo) |
| `node` version     | `...`                |
| OS and its version | `...`                |

#### Description
I've come across a bug when editing my entry: https://journaly.com/post/900
After getting some comments, I've partly edited it, saved it (let's call it version A), then returned to it, made some more changes and saved it again (version B). Now the text shown is version B. However, if I click edit again, the version in the editor is the version A. I've tried clearing the cache and also logging out and in again, but there's no change.

Later I edited on another device and this time it worked.

Robin wrote to me: "This is actually a bug we've had but it's been very hard to track down and reproduce as it doesn't seem to happen consistently. "

So below are the steps I can recollect, with no guarantee that it's possible to reproduce it

#### Repro Steps
Steps 1-2, 4-9 on desktop computer, step 10 on a mobile device, everything in the latest Firefox
I don't know if that matters...

1. User A creates a new entry and saves it as a draft (version 0.1)
2. User A updates the draft and saves it (version 1)
3. User B starts to comment some parts of the text
4. While User B is still making inline comments, user A edits parts of the text affected by the comments
5. User A saves his text (version 2)
6. User A now sees version 1 displayed with the comment of user B (the comment was not in the outdated section)
7. User A edits the text yet again and sees the correct version 2 in his edit window
8. User A saves the text again (version 3)
9. User A sees version 3 displayed, but when going into edit mode again, he sees version 2.
10. User A later loads the page on his mobile device and everything works correctly (version 3 shown and edited) Not sure whether the change of device or the timespan between steps 9 and 10 are the reason why the issue disappeared.

robin-macpherson:
Thanks so much for this @rickyweb ! We'll make some time to look into this and let you know if we need more information soon 🙏🏼

rickyweb:
@robin-macpherson, I've had the bug again today. It is not related to the comments as I've written yesterday. The issue is more likely to be located in the application cache (not the browser cache), the feature which will allow you to start writing something, and even without saving, going to another page inside journaly and then go back to your post without losing anything. That cache just needs to be deleted whenever one saves the entry.

robin-macpherson:
[...]

PR #447, opened 2021-01-13
Don't set a value when clearing autosave state.

Lanny:
## Description

Possibly addresses #416, it looks like when we save a post we "reset" the autosaved values, meaning clear the storage and restore the value to the initial value. The problem with this is the "restore the value to the initial value" step causes a state change and makes us write the reset value back to the local storage, and depending on how much clock drift there is between the server and client it might or might not be considered the more "recent" save. At least that's the theory.

@rickyweb if you have the time it would be awesome if you could try to repro #416 on the preview env.

### DIFF
--- a/hooks/useAutosavedState.ts
+++ b/hooks/useAutosavedState.ts
@@ -44,7 +44,6 @@ export default function useAutosavedState<T>(
 
   const reset = () => {
     delete storage[storageKey]
-    setValue(initialValue)
   }
 
   return [value, setValue, reset]


robin-macpherson:
Ohh this is interesting, curious to see if this indeed addresses the issue. It almost seems to simple if @rickyweb was able to reproduce this so reliably on the same post(s) over and over again. @Lanny do you see any risk with potentially merging this and letting @rickyweb see if it impacts things on the post he was having problems with?

Lanny:
No, I think we should be fine to merge. I believe what we'd see in prod is that the issue will happen the next time you edit a post, but then after saving it shouldn't happen again (if the hypothesis regarding the cause of the issue is correct)

robin-macpherson:
@Lanny got it, in that case I’m personally fine with merging to prod and trying that workflow with @ryanweb and maybe I can also reach out to LindasLinguas on Journaly because I know she had also been experiencing this issue as well. Does that sound good to you or would you prefer more testing on staging?

rickyweb:
I checked on preview and was not able to reproduce the faulty behavior. All local storage entries (except for the persistent one "welcome-modal-july-2020=seen") get cleanly cleared when I save a post.

Also, when leaving the edit-box to somewhere else and coming back to the post, edit it, it correctly loads the autosave values.

robin-macpherson:
@rickyweb great, thank you for verifying all of that! I'm going to merge this now and let's see if we can verify this on production!

robin-macpherson (approved):
🦔🤞🏼 A hopeful hedgehog! ✨

rickyweb:
@robin-macpherson , while all autosave entries _did_ disappear in preview when I saved a post, they don't in the prod environment. Or isn't it supposed to work just yet?

robin-macpherson:
> @robin-macpherson , while all autosave entries _did_ disappear in preview when I saved a post, they don't in the prod environment. Or isn't it supposed to work just yet?

Thanks for the update @rickyweb ! Just to be clear, are you testing on a new post in prod or your existing post that had the problem?

Is this taking into account @Lanny 's previous thought in the comments above?
> I believe what we'd see in prod is that the issue will happen the next time you edit a post, but then after saving it shouldn't happen again (if the hypothesis regarding the cause of the issue is correct)

rickyweb:
This was in an existing post, after having cleared the local storage. I edited the post, made a change and saved it. Then I went back to the post and got the version before my change (so I lost my change). Then I noticed the local storage had the fatal entry. So yes, I think that will count as taking into consideration @Lanny 's comment.

I'm going to try with a new post nevertheless. I'll let you know how it turns out.

robin-macpherson:
Ah, interesting. Sorry for all the hassle @rickyweb , we really appreciate your help testing this! For the new post, I wonder if we should test it thoroughly in staging so that we can freely leave you comments and make as many changes as we want without you having to write a legitimate post?

cc: @Lanny

rickyweb:
@robin-macpherson , no worries. It just took a minute to create a new test and I deleted the post immediately after.

So... It worked! I even recorded the thing as a screencast just in case it would fail again and I could send it to you. But now that it worked, no need any more! :)
cc: @Lanny

robin-macpherson:
### @rickyweb AMAZING!!! 🎉🎈 Agh this has been "bugging" me (cough) for so long! Thanks for the fix @Lanny , amazing how simple it was in the end

robin-macpherson:
So based on what we know now, what can we advise people to do who have reported experiencing this issue with their existing posts? You still had problems with your original post, right?

rickyweb:
Yes, that's right, my old posts still show this problem. I always have to delete the related autosave entries in the local storage of the browser in order to be able to work properly.

PR #450, opened 2021-01-13
Force skinny nav on settings pages

Lanny:
## Description

This is how it was intended to work before, but in the process of fixing the nav when pre-rendering I broke this. This isn't a super pretty kludge that involves rolling NavLinks into Nav, but I can't think of a better way to handle it. We could make it slightly cleaner by using a css preprocessor (SASS, LESS, whatever, it doesn't really matter) with `styled-jsx` but I don't really want to bring in the whole dep just for this.

## Migrations

No migs


components/Dashboard/Nav/Nav.tsx
```diff
@@ -32,14 +49,77 @@ const Nav: React.FC<Props> = ({ expanded, collapse }) => {
...
+    handleCollapse()
+    logout()
+
+    // Clear the currentUser cache
+    refetch()
+
+    // Redirect to home page
+    Router.push('/')
+  }
+
```
robin-macpherson, line 62: **NICE** thanks for fixing this!!

robin-macpherson (approved):
### WOOOOOOOOO THANKS FOR FIXING THESE! 💃🏻 🦔 💃🏻

components/Dashboard/Nav/Nav.tsx
robin-macpherson: Ah, @Lanny I just tested logout on staging and I'm seeing the same behavior as before. Super happy to get the nav fixed on settings page but just checking if this code was intended to also fix logout?

Lanny: It was not, this is probably the piece of code to look at but this PR was just trying to get the nav working on the settings page.

robin-macpherson: Ah ok cool will merge now then! 😄

PR #464, opened 2021-01-18
i18n / Translation Audit & German Translation Updates

### LINKED ISSUE #248: TRANSLATION AUDIT
#### Perceived Problem

- We have a number of places we're not using i18n / externalized translation strings
- We're also using the native Next.js router in some places

- Let's do an audit and get all this cleaned up before it becomes a really large task
- We'll need to switch out the native Next.js router for the next-i18next router

robin-macpherson:
## Description

**Issue:** closes #248 

There are a few places left in the app that aren't using externalized translation strings. This PR gets the last of 'em!

## Subtasks

- [x] Existing German translation fixes/updates
- [x] Add all new translation strings to UI Translation spreadsheet
- [x] FINALLY update landing page video!

**Add translation strings to:**
- [x] Update password screen
- [x] My Feed page
- [x] My Posts page
- [x] Post page
- [x] Settings page
- [x] Comments sections
- [x] Landing page

## Migrations

Nada


components/LanguageFormField/LanguageFormField.tsx
```diff
@@ -97,13 +99,13 @@ const LanguageFormField: React.FC<Props> = ({ languages, languageRelations, refe
...
-          placeholder={'Select a language...'}
+          placeholder={t('profile.languages.languageSelectPlaceholder')}
           options={langOptions}
           value={selectedLangId}
           onChange={setSelectedLangId}
         />
         <Select
-          placeholder={'Select a level...'}
+          placeholder={t('profile.languages.levelSelectPlaceholder')}
           options={levelOptions}
```
Lanny, line 109: The options for languages levels needs to get translated too. It's kind of a pain since they're in a function but you can either pass in the `t` function and let `fullLevelName` call it, or return a translation key from `fullLevelName` and call `t` outside..

Lanny (commented):
This is great! I had a folder of screenshots where we weren't translating the copy but it looks like you hit just about all of them. Just have one last piece of content that needs to go through translation, but good stuff!

Lanny (approved):

PR #467, opened 2021-01-20
Setup bundle analysis

Lanny:
## Description

Run the typical bundle analysis as part of the build step. Uploads an artifact. Next has been griping at us about bundle sizes lately so this seemed like a reasonable way to keep an eye on it and trim some fat if necessary.


Lanny:
Hey, it worked on the first try. Woot!

Example analysis: https://226-240991394-gh.circle-artifacts.com/0/.next/analyze/client.html

robin-macpherson:
Damn, I was hesitant to bring in Lodash but did it potentially temporarily back when I needed a deep object comparison - I wonder if we should remove it given how much of our bundle it's occupying. What do you think?

robin-macpherson (approved):
Epic! Great idea! 👌 😄

Lanny:
Yeah, I was surprised how much space lodash takes up. Given we don't use it heavily I'm fine with just dropping the dependency (or just getting it out of the FE code). Looks like the way lodash is packaged, we can change the way we import it to make it more tree-shakable: https://selleo.com/til/posts/0ac258eyv8-decrease-bundle-size-by-importing-lodash-correctly

But I'm kind of leaning towards just removing the dep and having one less thing to upgrade/manage vulnerabilities in/worry about bundle contribution

PR #473, opened 2021-01-22
Add .nvmrc, pin at node v14.15.4

Lanny:
Title pretty much says it all


robin-macpherson (approved):
Awesome thanks for knocking this out!

Issue #479, opened 2021-01-26
BUG: Impossible to write or edit Korean text on mobile


rickyweb:
|                    |                      |
| ------------------ | -------------------- |
| repro              | [link](https://todo) |
| `node` version     | `...`                |
| OS and its version | `android 6.x with Microsoft SwiftKey keyboard 7.7.0.4`                |

#### Description
When typing the Korean characters, one composes syllables letter after letter. So when writing 한국 (Korea), one types the following characters in this sequence: ㅎ,ㅏ,ㄴ,ㄱ,ㅜ,ㄱ. This has the following output:
ㅎ
하
한
한ㄱ
한구
한국

Now, when typing in the title box, everything works fine:
![Screenshot1](https://user-images.githubusercontent.com/2961791/105811953-a8265f00-5fad-11eb-9bc0-81583c46cd98.jpeg)

#### Repro Steps

However, when typing in the compose text field, things get very spooky, especially when you use backspace in order to delete a character.
1. Assume, you just wrote 하,
2. delete the ㅏ (in order to get back to just ㅎ, because you in fact wanted to write 호.) When deleting the ㅏ,
3. you get an error message:
![Screenshot2](https://user-images.githubusercontent.com/2961791/105811956-a8265f00-5fad-11eb-886e-939b033ff4ff.jpeg)
4. When you go back, you get this text:
![Screenshot3](https://user-images.githubusercontent.com/2961791/105811952-a78dc880-5fad-11eb-8412-d60cf4a37d15.jpeg)
5. And if you select everything and delete the whole selection, you get this output.
![Screenshot4](https://user-images.githubusercontent.com/2961791/105811949-a6f53200-5fad-11eb-912b-d999d4f30886.jpeg)

This makes it impossible to write in Korean.

As this is only happening with this web application, I believe the problem is not related to the mobile phone nor the keyboard used, but some "magic" happening when entering text, magic that works well in all other languages, but has these unwanted effects in a language with composed keystrokes like Korean.

Lanny:
Yeah :( ...

This is an upstream bug in slate, which is supposedly an upstream bug in chromium. It effects, AFAIK, every character composition based IME. Which is really really bad for a language learning app. It varies in severity depending on OS/browser it seems. We haven't had time/resources to go poking around in slate but that does seem to be where the fix needs to happen.

[...]

PR #484, opened 2021-01-28
Remove editor placeholder.

Lanny:
Possible partial mitigation for #479


robin-macpherson (approved):
Aww, I always felt fondly of this 😢 But if it turns out to be a fix, that would be huge

Lanny:
Unfortunately this didn't seem to fix anything, but we get to keep the the fun placeholder at least haha. Gonna close this since it didn't get us any milage

PR #485, opened 2021-01-29
Fill in last UI translation gaps!

robin-macpherson:
## Description

One last round of translation gaps hopefully! 🤞🏼 I've updated [the spreadsheet](https://docs.google.com/spreadsheets/d/1NkyPTvK4DCDtbSyMfhIj3pI8obhbC_CzHZJbyP19uEE/edit?usp=sharing) accordingly as well so everything should be ready for new language translations 💪🏼

## Subtasks

- [x] Update spreadsheet
- [x] Update code

## Migrations

No migs ✌🏼

### DIFF
--- a/components/JournalyEditor/helpers.ts
+++ b/components/JournalyEditor/helpers.ts
@@ -18,7 +18,7 @@ type ToggleByTypeArgs = {
   type: ButtonType
   editor: Editor
   format: string
-  t?: TFunction
+  t: TFunction
 }
 
 type ToggleArgs = Omit<ToggleByTypeArgs, 'type'>
--- a/components/JournalyEditor/helpers.ts
+++ b/components/JournalyEditor/helpers.ts
@@ -18,7 +18,7 @@ type ToggleByTypeArgs = {
   type: ButtonType
   editor: Editor
   format: string
-  t: TFunction
+  t?: TFunction
 }
 
 type ToggleArgs = Omit<ToggleByTypeArgs, 'type'>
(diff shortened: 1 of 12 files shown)


components/JournalyEditor/helpers.ts
```diff
@@ -140,7 +140,8 @@ const wrapLink = (editor: Editor, url: string) => {
 }
 
 const toggleLink = ({ editor, t }: ToggleArgs) => {
-  let url: string | undefined | null = window.prompt('Enter the URL of the link:')
+  const enterUrlPrompt = t ? t('enterUrlPrompt') : 'Enter the URL of the link:'
```
Lanny, line 143: hmm, what is the situation where `t` would be falsy?

robin-macpherson: Yeah I did this for now to satisfy TS because `t` was set as optional when created, but I agree that if things are set up properly it shouldn't ever be falsy. Shall we just make it required so we can remove this check?

Lanny: Yeah, I think making it required in `ToggleByTypeArgs` makes sense. I'm guessing it was marked optional for a reason so we may have to move a null check up to where we call into `toogleByType` (heh, just noticed that misspelling) but we almost certainly should be checking it down here in these toggle functions

Lanny (approved):
One comment, but I think we're on the same page about what should be done so approving this. Good stuff!

PR #486, opened 2021-01-31
Avarar specific sizing

Lanny:
## Description

Adds support for multiple transform configurations, our second configuration is for avatar images. Previously we were reusing the post image config so we'd get a 1200px by something image which was wayyyy bigger than we'd ever present to the user. Now avatar uploads are 150px square. This should reduce the total network IO required to load the feed and comments, plus save us a little on S3 costs. Main goal here however is a PoC for multiple image transforms which will be necessary for inline images.

## Migrations

No migs

Related thumbbuster PR that needs to be deployed first: https://github.com/Journaly/thumbbuster/pull/1


robin-macpherson (approved):
Epic epic epic 🎁... 🦔💨🔥

Issue #482, opened 2021-01-28
Can't upload avatars/profile images


Lanny:
Original report:

```
Hi Robin,
I was looking for where to ask technical questions/report anything technical on Journaly.
Could you tell me where I should send that kind of messages please?
Btw, it's nothing big, don't worry. Just been unable to upload a profile picture
(I've tried while bejng connected to different wifis, tried several file sizes down to rather small but it keeps loading forever so I thouht maybe something is wrong with that feature) .
Btw, I am LOVING Journaly!!!!!!!!!!!!
```

Probably a result of thumbbuster changes.

Lanny:
@candacepowell @robin-macpherson I wasn't able to repro this on prod. I tried on chrome and firefox on OSX. Can we reach out the folks reporting this issue and find out their OS/browser combos, and if we could get one of the images they're having trouble uploading that would be awesome

candacepowell:
@Lanny @robin-macpherson - Done. Will let you know once I receive info back.

candacepowell:
For what it is worth, I couldn't replicate this today either. 

macOS Catalina v10.15.7
Chrome Version 88.0.4324.96

**Steps to Reproduce:**
1. https://journaly.com/dashboard/settings/profile
2. Click "Upload Image"
3. Select file (see attached below)
4. Loading circle appeared for max 5 seconds
5. Notification received that image upload was successful (blue sliding bar)
6. Image saved on profile

![Me](https://user-images.githubusercontent.com/77520576/106337287-e93bad80-62e4-11eb-9d87-0e6fb4e753e3.jpg)

candacepowell:
We have a reply from the user. 🙌🏼 

Chrome
Windows 10

_Images:_
![AVA-IslaMujeres](https://user-images.githubusercontent.com/77520576/106367867-e0f17a00-6399-11eb-97d9-e48d90266c36.jpg)
960 × 720, 45KB

![AVA-Learn ProfilePic](https://user-images.githubusercontent.com/77520576/106367913-47769800-639a-11eb-94ad-26a8e3ae224c.jpg)
511 × 515, 37KB

And they've also supplied two others which I can only save as .txt for some reason? But all 4 were unsuccessful.

robin-macpherson:
Thanks for getting these @candacepowell ! 🙌🏼

robin-macpherson:
Two posts that users have written publicly about this issue:
- https://journaly.com/post/3427
- https://journaly.com/post/3489

[...]

PR #487, opened 2021-01-31
More robust error handling in image uploading.

Lanny:
## Description

**Issue:** #482

Catch exceptions when doing XHR, hopefully will make whatever issue people are running into with #482 more apparent and at least won't get them stuck in an infinite spinner state.


robin-macpherson (approved):
Great stuff, thanks for the quick turnaround mate!

PR #493, opened 2021-02-06
Use custom slate build with some patches

Lanny:
## Description

**Issue:** #479 (possible/partial mitigation)

This bug really sucks and is embarrassing to have as a platform dedicated to foreign language learning. Every bug we've found with IME so far has been reproducible upstream, so not strictly our fault, but users probably don't care. I ended up forking slate and am maintaining a new branch here: https://github.com/Journaly/slate/tree/journaly-patch . This branch will go away eventually

Two changes to that branch are:
- Include this PR: https://github.com/ianstormtaylor/slate/pull/3698 which fixes issue with cursor/selection movement when using IME on safari (mobile and desktop) (not sure if this has any impact for #479)
- Remove placeholder more definitively: https://github.com/Journaly/slate/commit/de49a321fbec73e4d621e7726f09975d4b06e770 which should fix the situation where the whole app crashes on IME entry at the start of the file (this should fix the crashing behavior detailed in #479)

## Migrations

FE only, no migs


robin-macpherson (approved):
Ahhh amazing! Thanks so much for doing this mate, these bugs are definitely the main ones remaining that keep me up at night since they basically destroy the writing experience for users that experience them and it sucks that they are nothing to do with our code, and also the fact that it's very difficult to properly convey this to our users. Not to mention, they shouldn't really have to care, all they see is a bad experience and I could definitely see this turning some users away pretty quickly. So yeah it's just great to see some progress here 🙏🏼🔥

PR #497, opened 2021-02-09
Journaly Project Restructure - Feb 2021

robin-macpherson:
## Description

**Issue:** #400 

The Journaly project has come a LONG way! 🔥🚀

We now have multiple services with separate deployments leveraging a lot of our own Cloud Infrastructure. We've learned a lot about our tools and technologies, and also some of the technologies we use have grown and matured a lot, too (Prisma & Vercel).

We believe that it is important to take time to reflect on the evolution of the codebase and make intentional and thoughtful updates as this happens to keep things clean, tidy, well-organized, and pleasant to work within, and not allow things to grow into a tangled mess that eventually becomes a nightmare to fix.

So with that philosophy in mind, this is an effort to restructure the whole project so that it has clear and logical separation of concerns and is a lot more intuitive to navigate. See README updates (TODO) for a complete overview.

## Subtasks

- [x] Move everything around
- [x] Update `package.json`s and generate new lock files
- [x] Update alias paths and all imports
- [x] Update imports for merged `elements` & `components`
- [x] Run TS type-check on everything
- [x] Update & test Vercel config
- [x] Get CircleCI working

## Migrations

No Migs

## Screenshots

☕️  ☕️  ☕️  ☕️  ☕️  ☕️  ☕️
☕️   ☕️  ☕️  ☕️  ☕️  ☕️ ☕️
☕️  ☕️  ☕️  🧘🏻‍♂️  ☕️  ☕️  ☕️
☕️  ☕️  ☕️  ☕️  ☕️  ☕️  ☕️
☕️  ☕️  ☕️  ☕️  ☕️  ☕️  ☕️


robin-macpherson:
@Lanny in your original proposal you used the term `packages`, just wondering what you think about using `services` instead? So `journaly/services/web` for example

Lanny:
No strong feelings, the one place I think "services" doesn't quite work is j-db-client which is closer to a library consumed by web and j-mail than a standalone service

robin-macpherson:
> No strong feelings, the one place I think "services" doesn't quite work is j-db-client which is closer to a library consumed by web and j-mail than a standalone service

Ah, hmm yeah that's a good point. In the interest of keeping things as simple as possible I'm fine with just sticking with `packages` then ☮️

Lanny (approved):
woot!

PR #498, opened 2021-02-09
Update project documentation - Feb 2021

robin-macpherson:
## Description

Following our big project re-org in #497 , this PR is to clean up, update, and improve our documentation 🎉

## Subtasks

- [x] Update main README
- [x] Make sure each service/package has some docs

## Migrations

Nada

## Screenshots

Marvel at those beautiful READMEs

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Here is an overview of the project setup:
 |------ web/ # Our core application that users engage with on the web
 |------ j-mail/ # A simple managed queue service for handling transactional emails from within the app
 |------ j-db-client/ # A database client that can be imported into any service to interact with the db
-|-------- datamodel.prisma # our database schema
+|-------- schema.prisma # our database schema
 |-------- migrations/
 ```
 
--- a/README.md
+++ b/README.md
@@ -2,21 +2,47 @@
 
 ###### @ [journaly.com](http://journaly.com)
 
-_A foreign-language journaling application where you can exchange feedback with native speakers._
-
----
-
 Welcome to the Journaly repository!
 
-The application is built with JavaScript/TypeScript on both the frontend and backend with a GraphQL API, leveraging powerful frameworks such as React, Next, and Apollo on the frontend and Prisma, Nexus, and Express on the backend, with a PostgreSQL database with Amazon RDS.
+_A foreign language journaling platform for exchanging feedback with native or advanced speakers and building community around shared interests and languages._
 
-We use Python for data science, leveraging a number of powerful libraries for data aggregation and analysis, data visualization, and machine learning, such as NumPy, Pandas, Matplotlib, Seaborn, SciKit, and more.
+## Why?
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -27,19 +53,20 @@ The data model/DB schema lives in `j-db-client/prisma/schema`, server side code
    $ git clone git@github.com:Journaly/journaly.git
    ```
 1. Locate the `.env.example` file and copy the contents into a new `.env` file alongside it.
+1. Navigate to `packages/web`
 1. Run `npm ci`
-1. Run `npm run prisma:generate`
```
robin-macpherson, line 31: @Lanny removing this step because it should happen with the prisma `post-install` hook, right?

README.md
```diff
@@ -2,21 +2,47 @@
...
+|---- cypress/
+|---- packages/
+|------ web/ # Our core application that users engage with on the web
+|------ j-mail/ # A simple managed queue service for handling transactional emails from within the app
+|------ j-db-client/ # A database client that can be imported into any service to interact with the db
+|-------- datamodel.prisma # our database schema
+|-------- migrations/
+```
+
+More detailed documentation can be found within each directory (WIP).
```
robin-macpherson, line 45: @Lanny I went back-and-forth between this simplified version or adding some additional depth below each package. For this first pass I decided to keep it very simple just as a high-level overview, and then directing people to the specific docs for each package if they need more info. Do you have any thoughts?

robin-macpherson:
Just FYI @JoshuaBelden ⬆️ Hopefully these updates are a big improvement on the docs overall, and should also be helpful with your Docker efforts 🤞🏼

README.md
Lanny: This approach seems reasonable to me. I might even go so far as to say even this breakdown isn't necessary here. Stuff like cypress and docs should be pretty explanatory, and a breakdown of what each package does might be more suitable for packages/readme.md (and we'd have room to expand a little bit on each there)

README.md
Lanny: In theory yes, I think we needed to add it to the `build` npm script because circle was restoring node_modules from a cache and not re-running it? Actually that brings up a good point that you need to run `npm run build` at least once for nexus reflection and all the static analysis stuff that kids love these days to work, so that would probably be good to add here

README.md
```diff
@@ -27,19 +53,20 @@ The data model/DB schema lives in `j-db-client/prisma/schema`, server side code
...
 1. Locate the `.env.example` file and copy the contents into a new `.env` file alongside it.
+1. Navigate to `packages/web`
 1. Run `npm ci`
-1. Run `npm run prisma:generate`
 
 #### Setting Up Your Local DB Instance
 
 1. Install Postgres
+
 - Mac/Linux users should use Homebrew
```
Lanny, line 63: Linux users should _probably_ be using their distro's postgres package manager rather than linux homebrew

Lanny (approved):

Issue #496, opened 2021-02-08
Bust the cache after creating a new post


robin-macpherson:
#### Perceived Problem

- Currently, in production, when a user creates a new post and then visits there "my posts" page, they don't see that new post because of our Apollo cache.
- This is causing some users to think that we lost their post completely. In some cases the user ends up creating multiple duplicate posts trying to figure out what's happening.

#### Ideas / Proposed Solution(s)

- We should bust the Apollo cash whenever someone creates a new post to ensure they see the updated data when visiting the feed or my posts page

davidsonsns:
Hi guys! Finally, I'm back!
I'd like to get this one. Could you assign it to me? Sounds that it won't take so long to be fixed, with that I can help with anything else after fixing an in-production bug.

robin-macpherson:
Thanks @davidsonsns ! 🙏🏼 I've assigned it to you. Are you still planning to take #434 ?

davidsonsns:
Lol! For sure I can do it, definitely forgot it. 😒

Thanks for reminding me!

robin-macpherson:
lol awesome thank you @davidsonsns ! 🎉 By the way, I wanted to make you aware of this big project re-org just went in finally! 🎈 #497 

I'm going to be working on payments infrastructure next and Lanny is working on inline post images (our first premium feature). I do have a really exciting project for you next if you have more time after the current things 😄 🔥

davidsonsns:
Of course, I'm game! Thanks for your feedback about all these new features.
That sounds great! I'll be very happy to keep contributing with you guys on the project.

robin-macpherson:
Wonderful, it's been really great to have your help! 😄 So the next project I was thinking could be great for you after this and #434 is **Adding markdown support for commenting:** #425 – that one is going to be a HUGE improvement to the app and user experience! Let me know if it sounds like something you'd be excited to work on 😀

davidsonsns:
Yeah! That sounds great @robin-macpherson.
Just saw your amazing work related to the structure. I have some points that could improve it even more (at least I think so 😇). I'll create a PR related to it ASAP.

PR #500, opened 2021-02-09
fix(496): add created post on cache

davidsonsns:
## Description

**Issue:** #496 

After creating a post as a draft or publishing it, the user will be able to see this one on the list of `My posts`

FYI: Instead of invalidating the query, I'm following the previous implementation of update the cache after receive a request success callback. With this approach, we can avoid an extra request.

## Subtasks

component `packages/web/pages/dashboard/new-post.tsx` refactors:
- [x] use `toast` in case of error;
- [x] remove sending generic types on `useState`, once the initial value can be an empty string;
- [x] use `useCreatePostMutation` callbacks;

### DIFF
--- a/packages/web/pages/dashboard/new-post.tsx
+++ b/packages/web/pages/dashboard/new-post.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef, useState } from 'react'
 import { NextPage } from 'next'
 import { useRouter } from 'next/router'
 import { withApollo } from '@/lib/apollo'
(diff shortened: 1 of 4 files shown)


packages/web/pages/dashboard/new-post.tsx
```diff
@@ -15,10 +15,14 @@ import {
...
   useCreatePostMutation,
   PostStatus as PostStatusType,
+  PostsQuery,
+  PostsQueryVariables,
+  PostsDocument,
 } from '@/generated/graphql'
 import AuthGate from '@/components/AuthGate'
 import { useTranslation } from '@/config/i18n'
 import useUILanguage from '@/hooks/useUILanguage'
+import { toast } from 'react-toastify'
```
robin-macpherson, line 25: Let's put this library import at the top

robin-macpherson (approved):
Amazing stuff @davidsonsns , love the approach of updating the cache and saving that extra request! 💪🏼🔥 Just one very minor request from me but looks great!

packages/web/pages/dashboard/new-post.tsx
davidsonsns: created the pr #501 following the approach that I've been using so far

Lanny (approved):
LGTM 🎉

Issue #434, opened 2021-01-08
Use Optimistic UI for giving "thanks" on comments


robin-macpherson:
#### Perceived Problem

- the action of giving thanks for a comment is a little slow and can result in the user clicking it multiple times

#### Ideas / Proposed Solution(s)

- I think we should use "optimistic UI" here, meaning that we assume the operation was successful and update the UI immediately in order to have a better user experience and prevent confusion
- We can always display an error if for some reason the operation failed

davidsonsns:
It'll be good to remember some apollo concepts. Could you assign it to me?

robin-macpherson:
@davidsonsns absolutely! Thanks for taking it 😄🙏🏼 By the way, just to make you aware I will hopefully be completing a pretty big project re-org this weekend. It shouldn't be too bad hopefully since the PR for this issue should be quite simple but I just wanted to let you know 🙂

davidsonsns:
That sounds amazing!

Don't worry I'll send a small one. Anyway, I was thinking about applying this same optimistic behavior to update and delete features. That sounds good? If so, I'll split it into more than one or. It will help with the reviews.

Is this refactor related to PR #400? Just saw it here right now. It'll definitely be helpful.

robin-macpherson:
Yep it's based on the proposal in #400 😄

Regarding optimistic UI for updates and deletes, are you specifically just talking about for comments?

davidsonsns:
Yeah, just about post comments

robin-macpherson:
Ok sounds good! 💪🏼

davidsonsns:
@robin-macpherson can it be closed?

robin-macpherson:
@davidsonsns absolutely! Sorry for the delay 🙏🏼

PR #503, opened 2021-02-11
feat(434): use optimistic thanks on comments

davidsonsns:
## Description

**Issue:** #434 

It will add an optimistic UI experience to the thanks on comments.

## Subtasks

- [x] refactor types
- [x] memorize the value of the variable `hasThankedComment`
- [x] apply progress cursor when the `thanks` action is loading

> FYI: Just made a re-research about the approach to update the cache data on the latest apollo client, was very unhappy with the last approach I used before using `cache. readQuery` and `cache. writeQuery` (used to use it in my previous experience with apollo-client ). Anyways, found this function `cache.modify` that definitely is very easy to implement and remove all that complexity adopted before.
> - https://www.apollographql.com/blog/first-impressions-with-apollo-client-3-2ae2a069ab2f/
> - https://www.apollographql.com/docs/react/caching/cache-interaction/#cachemodify

## Screenshots
https://www.loom.com/share/2a4478d009304e5496193d7f3e77a45e

### DIFF
--- a/packages/web/components/InlineFeedbackPopover/Comment.tsx
+++ b/packages/web/components/InlineFeedbackPopover/Comment.tsx
@@ -1,5 +1,7 @@
-import React, { useState, useRef } from 'react'
+import React, { useState, useRef, useMemo } from 'react'
 import Link from 'next/link'
+import { toast } from 'react-toastify'
+import classNames from 'classnames'
 
 import {
   useUpdateCommentMutation,
@@ -7,9 +9,10 @@ import {
   CommentFragmentFragment as CommentType,
   useCreateCommentThanksMutation,
   useDeleteCommentThanksMutation,
+  CommentThanks,
+  UserFragmentFragment,
 } from '@/generated/graphql'
 import { useTranslation } from '@/config/i18n'
-
 import Button, { ButtonSize, ButtonVariant } from '@/components/Button'
 import BlankAvatarIcon from '@/components/Icons/BlankAvatarIcon'
 import theme from '@/theme'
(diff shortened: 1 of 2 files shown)


packages/web/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -69,35 +72,75 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment, cu
...
+              )
+            },
+          },
+        })
 
-  const [deleteCommentThanks] = useDeleteCommentThanksMutation({
-    onCompleted: () => {
-      // just refetches the post as in updateComment
-      onUpdateComment()
+        cache.evict({ id: `${data.deleteCommentThanks.__typename}:${data.deleteCommentThanks.id}` })
```
robin-macpherson, line 123: lol "evict" makes me chuckle 😆

robin-macpherson (approved):
Looks good to me!

packages/web/components/InlineFeedbackPopover/Comment.tsx
davidsonsns: 😁

packages/web/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -69,35 +72,75 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment, cu
...
-        },
-      })
+      deleteCommentThanks({ variables: { commentThanksId: currentCommentThanks.id } })
     }
   }
 
+  const isLoadingCommentThanks = useMemo(
+    () => createCommentThanksResult.loading || deleteCommentThanksResult.loading,
+    [createCommentThanksResult.loading, deleteCommentThanksResult.loading],
+  )
```
Lanny, line 142: I'd just do
```const isLoadingCommentThanks = createCommentThanksResult.loading || deleteCommentThanksResult.loading```

useMemo is going to have to do at least two comparisons and a conditional branch, which is already going to be slower than just ORing two values and function creation and call overhead adds to that too. It's a boolean so it's not like memoizing it is necessary to prevent re-renders when passing it along as a pro

File: packages/web/components/InlineFeedbackPopover/Comment.tsx [hunk 3]
```diff
@@ -281,6 +324,12 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment, cu
         .like-btn :global(svg:hover) {
           cursor: pointer;
         }
+        .progress {
+          cursor: progress;
+        }
+        .pointer-events-none {
+          pointer-events: none;
+        }
```
Lanny, line 332: This is kind of nitpicky, but you can save yourself the extra class with something like:

```suggestion
        .progress > .like-btn {
          pointer-events: none;
        }
```

packages/web/components/InlineFeedbackPopover/Comment.tsx
```diff
@@ -183,10 +226,10 @@ const Comment: React.FC<CommentProps> = ({ comment, canEdit, onUpdateComment, cu
...
       )}
-      {!canEdit && currentUserId && (
-        <div className="edit-block">
+      {!canEdit && currentUser?.id && (
+        <div className={classNames('edit-block', { progress: isLoadingCommentThanks })}>
           <span
-            className="like-btn"
+            className={classNames('like-btn', { 'pointer-events-none': isLoadingCommentThanks })}
             onClick={hasThankedComment ? deleteExistingCommentThanks : createNewCommentThanks}
           >
```
Lanny, line 234: Know you didn't do this, but this button should be a button element rather than a span, or at least have role=button if it has to be a span. If you don't mind changing it while we're here that would be awesome, otherwise we can start a misc a11y ticket and add it as a todo there.

Lanny (approved):
Good stuff! You are our apollo cache wrangling wizard.

packages/web/components/InlineFeedbackPopover/Comment.tsx
davidsonsns: Lol! You're completely right. Sometimes I'm very addicted to memoization.
Thanks for reviewing it.

packages/web/components/InlineFeedbackPopover/Comment.tsx
davidsonsns: that's a good one! 💪🏼

packages/web/components/InlineFeedbackPopover/Comment.tsx
davidsonsns: As a developer, we're frequently forgetting about it. Thanks for this one! Really appreciate this kind of review.

I'll follow with `role=button` to avoid extra CSS code removing button style here. But, it will be great to create another ticket to revisit accessibility in the project. May discuss and define a pattern to follow.

davidsonsns:
> Good stuff! You are our apollo cache wrangling wizard.

@Lanny
LMAO! Leaning things every single day 😅

Issue #517, opened 2021-03-06
[web] - duplicated dependencies


davidsonsns:
I just ran the application with `yarn`(an automatic action, because of my work where we're using it) and saw this.

![Screen Shot 2021-03-06 at 20 44 45](https://user-images.githubusercontent.com/9276511/110224770-4e268c80-7ebd-11eb-88a7-3c944075a3d4.png)

robin-macpherson:
Thanks for this one @davidsonsns ! 🙏

PR #518, opened 2021-03-06
chore(517): fixing duplicated web dependencies

davidsonsns:
**Issue:** #517


davidsonsns:
Agree with you and I'm adopting this approach also. Didn't move all `@types` to make it small as possible and to fix a specific warning. Anyways, it could be turned into another PR to just make this chore.

PR #522, opened 2021-03-11
Add GA

Lanny:
## Description

Everyone's least favorite analytics platform! Well at least it wasn't a lot of code. 

`NEXT_PUBLIC_GA_ID` needs to be set for relevant envs.

Also fixed the issue where logging out wouldn't visibly take effect until the user refreshed (since it was annoying when testing GA for login)

## Migrations

No migs

Also fixes minor layout issue for non-auth'd users with nav:

Before:
Screen Shot 2021-03-11 at 4 35 40 PM

After
Screen Shot 2021-03-11 at 4 35 00 PM

### DIFF
--- a/packages/web/components/Dashboard/Nav/Nav.tsx
+++ b/packages/web/components/Dashboard/Nav/Nav.tsx
@@ -124,9 +124,11 @@ const Nav: React.FC<Props> = ({ expanded, collapse, disableLargeNav }) => {
         )}
         
         <div className="nav-support">
-          <span className="help-btn" onClick={() => setShouldShowModal(true)}>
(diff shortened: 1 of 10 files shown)


Lanny:
@robin-macpherson this is ready for review. Vercel is configured with the new env var so can be merged if you're good with the changes. Also includes a fix for SSR that seems to have been broken during the re-org, although may have been upstream shenanigans in Apollo and just ended up in that issues because it changed deps and re-generated package-lock.json.

robin-macpherson (approved):
Woot! Awesome thanks @Lanny 🚀

Issue #513, opened 2021-03-02
User's cannot open highlights on UNDERLINED text


robin-macpherson:
#### Description

A user noticed in [this post](https://journaly.com/post/4620) that clicking on highlighted underlined words doesn’t open the correction box, so if a highlighted string of words has no unhighlighted words, there’s no opening it. 

For instance, the line `lleva una vida estudiantil y personal bastante exitosa.` won’t open unless I click on the period, which isn’t highlighted.

#### Repro Steps

1. Go to https://journaly.com/post/4620
2. Try to click on any highlights directly on underlined text and see they don't open
3. For highlights containing a mix of underlined and normal text, you can still click on the highlight on normal text

PR #528, opened 2021-03-29
More misc fixes

Lanny:
## Description

**Issue:** #513 

Also fixes issue where newlines in post comments and thread comments are not rendered, which I don't think we have a ticket for.


robin-macpherson (approved):
Awesomeeee, thanks mate!

Issue #380, opened 2020-11-09
In-App Notifications


robin-macpherson:
#### Perceived Problem

- Currently, our only way to notify users of important activity, such as receiving feedback, is through email.
- This results in a lot of emails
- It can be tricky/inconvenient for users to stay up-to-date with everything that has happened since they last logged in

#### Ideas / Proposed Solution(s)

- Build an in-app notification feed where users can easily get a snapshot of everything they need to know and have a single source of truth

##### Notifications

- thread comment
- post comment
- badge/milestone achieved

##### Mobile

I'm thinking we could use repurpose this space where we currently show "Journaly" to display a notifications icon along with the number of unread notifications which would open the notifications page:

<img width="400" src="https://user-images.githubusercontent.com/34203886/98608459-15c0ca00-22a0-11eb-880f-07b0d3046c56.jpg" >

##### Tablet/Desktop

**Idea 1**

My original wireframes for the dashboard include a nice section for stats (day streak, thanks received or posts reviewed, and posts written). One idea is that we could build that out and somehow use that space/area to also show a notifications icon 🤔

![image](https://user-images.githubusercontent.com/34203886/98608674-9089e500-22a0-11eb-9a66-ed3c9ed263f3.png)

**Idea 2**

Perhaps the simplest idea would be to simply add a third dashboard icon under the separator for notifications. This should be easy to pop in there, for now it can just go to a dedicated notifications page to keep things easy and simple, and this leaves that other space for the stats bar mentioned in idea 1

![image](https://user-images.githubusercontent.com/34203886/98608892-05f5b580-22a1-11eb-9abe-1bad9da5d9fd.png)

robin-macpherson:
@Lanny I've updated this with a few ideas. I'm thinking that I prefer idea #2. It feels like a very nice and simple solution! I guess that would even negate the need for the suggestion about mobile, although I still think it could be good to use that top bar area more effectively with an additional notifications entry point and maybe even some stats or something

PR #529, opened 2021-03-31
Simple Profile Stats

Lanny:
## Description

Adds some stats to the profile page. Current not super visually appealing

## Migrations

No migs

## Screenshots

Screen Shot 2021-03-31 at 2 47 46 PM

@robin-macpherson here's what I have right now, could definitely use some design input on how to make the "summary stats" section prettier, and also general input and opinions on using the toggle tabs here.


robin-macpherson (commented):
Awesome idea, love the little calendar view! Maybe in the future we can even have one colour for wrote a post, one for gave feedback, and one for did both? I really like the colours on there now!

One thought I had which is just purely a fun thing around what language to use is maybe instead of "Member since" we could use "Journaling since"? Again not a big difference just an idea.

In terms of visual design, I think one thing that would help is to tighten up the horizontal space a little, so perhaps even centring the stats, putting "Total Posts" and "Thanks Received" side-by-side with "Journaling since..." on the next line (and then on mobile devices they can wrap to 1 stat per line). Another option would be just to have them aligned with the left of the calendar, which would also tighten things up nicely.

I'm only just seeing that the user's name on the left side doesn't have the same margins/spacing as the other info on the page, so if it's super quick & simple maybe we could also make the user's name line up on the left with their languages, interests, and bio?

Personally I think that with these super small changes this could totally go out and be a nice enhancement to the profile page! And the idea about diff colours for posting/feedbacking/both can be something we add later if you like the idea!

Lanny:
Awesome, thanks for the feedback!

>Maybe in the future we can even have one colour for wrote a post, one for gave feedback, and one for did both? I really like the colours on there now!

Totally, my thought there was to have checkboxes or maybe radio buttons at the bottom that would let you pick between like "Posts", "Feedback" and "Both" to be shown on the calendar view instead of color coding, just because the coloring is like a sliding scale so I'm not sure the best way to turn it into a 2D scale.

>One thought I had which is just purely a fun thing around what language to use is maybe instead of "Member since" we could use "Journaling since"? Again not a big difference just an idea.

I like that, it's a nice little touch that makes the copy less sterile, I'm all for it 👍

Lanny:
@robin-macpherson fixed the issue with the profile name. I kind of got inspired by the "journaling since" comment and thought it might be cool to have a little dynamically generated paragraph instead of a list or table of stats, sorta seemed to go with the theme of writing. I think it's kind of fun but totally understand if it seems corny. This is what it looks like:

Screen Shot 2021-04-05 at 8 40 19 PM

Some issues with pluralization but I figured I could fix that when I do a translation pass (thinking I'll do it in followup PR since profile stats aren't highest priority on our todo list)

Also note that the stats break if a user has a null create time on a published post, which is a situation that only happens with the test users, but just in case you want to look at it locally or in stage just know it's not actually broken for real users.

robin-macpherson:
Looks awesome @Lanny !! The only small comment I have is maybe instead of "inline comment" we can call it "inline feedback comment" or something just to make sure people know exactly what we're talking about. Can't wait to see this go live, it's a splendid little enhancement I think! We can also re-use some of this for the designs I had for displaying the stats on the dashboard feed page (see the designs I posted in #380  for example)

robin-macpherson (approved):
Woot wooooot! 🦔🔥

PR #533, opened 2021-04-07
Remove margin between paragraphs

### LINKED ISSUE #531: Paragraphs render inconsistently between editor and post
Paragraphs in a post when you're reading it get a margin that isn't applied in the editor. This leads to situations where a user will enter two new lines (creating an empty paragraph) between two paragraphs to provide appropriate space between paras in the edtior, however when the post is rendered for readers, the space between paragraphs will be significantly greater than what they say in the editor.

Two obvious ways to fix this:

1. Apply the margin between paragraphs in the editor, so the editor accurately reflects what a published post looks like today. If we take this route, we should probably bump the amount of margin applied in both cases as paragraphs without an empty line between them look a little crowded today.
2. Remove the margin in the rendered post, again making the two consistent.

Personally I'm in favor of 2. since I'm accustomed to adding a blank line between paragraphs as a general habit when typing, editors assuming how I want to space paragraphs (or assuming every newline means new paragraph) kind of annoy me, but that might just be a me thing. This also gives a little more flexibility as it allows a post author to put two lines right on top of each other by not putting a empty line between them, if they want to do that for some reason (perhaps when enumerating things in a list, or if they want to push content very close to an inline image or table).

The downside of 2. is that folks that are not in the habit of spacing paragraphs out with an empty line will end up with tightly packed paragraphs.

Lanny:
Fixes #531

### DIFF
--- a/packages/web/components/Dashboard/Post/Post.tsx
+++ b/packages/web/components/Dashboard/Post/Post.tsx
@@ -561,10 +561,6 @@ const Post = ({ post, currentUser, refetch }: IPostProps) => {
           background-color: ${theme.colors.highlightColorHover};
         }
 
-        .post-body p {
-          margin: 10px 0;
-        }
-
         .post-body table {
           width: 100%;
           margin-bottom: 10px;


robin-macpherson (approved):
Weehoooooooo ✨

PR #535, opened 2021-04-07
Confirm when deleting comments

### LINKED ISSUE #441: CONFIRMATION MODAL: Comment deletion
#### Perceived Problem

- Currently, if a user accidentally clicks "delete" on their comments, it is immediately deleted
- Some users have reported accidentally doing this

- Let's add an "Are you sure you want to delete this comment?" confirmation modal when someone clicks the delete icon
- Include both ThreadComments and PostComments
- We have an existing pattern for this with post deletion, see usage of `useDeletePostMutation` to find it

Lanny:
Fixes #441

Also add simple shortcut ctrl+enter to submit comments when typing int comment textarea

### DIFF
--- a/packages/web/components/Button/Button.tsx
+++ b/packages/web/components/Button/Button.tsx
@@ -28,6 +28,7 @@ type Props = {
   className?: string
   disabled?: boolean
   loading?: boolean
+  title?: string
   id?: string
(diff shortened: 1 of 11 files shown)


robin-macpherson (approved):
🔥🔥🔥🔥🔥

PR #502, opened 2021-02-09
💰🏗 Build out Payments Infrastructure

robin-macpherson:
## Description

**Issue:** closes #328 

Builds out the payments infrastructure for Journaly Premium as well as the UI flows for:
- subscribing to premium
- upgrading/downgrading membership
- cancelling membership
- cancelling the cancellation of membership

Involves pretty significant setup with a webhook endpoint to handle Stripe Webhook events.

Also includes some cleanup with how we instantiate `PrismaClient` instances.

## Subtasks

- [x] I have added this PR to the Journaly Kanban project ✅
- [x] Scaffold out the UI for everything
- [x] Set up Journaly Stripe account
- [x] Set up Stripe Webhooks & an endpoint to handle their events
- [x] Figure out how the data model should look and understand Stripe Subscription life-cycle and how that plays into our flows...
- [x] Get happy path working!
- [x] Handle payment failure (super basic for now)
- [x] Get cancellation working
- [ ] ...

## Migrations

This PR contains significant migrations that also need to be applied in a particular sequence with some custom SQL being run in-between.

- Does the migration need to be applied before or after deploying code?
- Is applying the migration going to necessitate downtime?
- Is there any SQL or backfill logic that has to be run manually?
- Are the DB changes high risk in your estimation? Should a manual DB snapshot be taken before applying?

PR #539, opened 2021-04-12
Logging woo!

Lanny:
Ignore this, just creating it for the preview env.


robin-macpherson:
@Lanny just double-checking this can be closed right, since we brought the changes into #502 ?

Lanny:
Yup yup, this branch is dead now

PR #546, opened 2021-04-23
Next i18next upgrade

### LINKED ISSUE #543: Upgrade next-i18next
As of writing we're two major releases behind. Nothing is broken but this error we get a dozen times on every build:

```
18:43:50.519 | Read more: https://err.sh/next.js/opt-out-auto-static-optimization
-- | --
18:43:50.519 | Warning: You have opted-out of Automatic Static Optimization due to `getInitialProps` in `pages/_app`. This does not opt-out pages with `getStaticProps`
```

appears to be the result of their HoC and the current version does not seem to have the same issue. Think this should be a fairly straight forward upgrade, but possible there's some API change in there that could cause problems. Just smoke testing the German translation should be fine.

Lanny:
## Description

Fixes #543

Upgrades next-i18next. Requires a pretty horrific hack because next-i18next doesn't support use of `getInitialProps` anymore, even though we're tied to it for how we do SSR with apollo (for now). Ultimately we'll want to move all our initial GQL calls into `getServerSideProps` instead of using apollo's `getDataFromTree` approach, but this has performance implications for client side caching (basically makes the client side cache useless between pages, hopefully next and apollo figure out how they're supposed interoperate on this by the time we're forced to upgrade).

Also this is a fairly far reaching change, so I'd like to test it out on stage a little bit before pushing it to prod. Meaning it's blocked by payments. That shouldn't be far off though so just going to leave this as a draft for now.

## Migrations

No migs


robin-macpherson:
@Lanny I propose we close this PR, since we've decided to drop this dep as soon as possible?

PR #548, opened 2021-04-26
Inline image misc fixes

### LINKED ISSUE #547: Journaly Editor click target is too small
#### Description

Currently the click target within the Journaly Editor is just the first line, so you have to click in quite a specific and small area to get the cursor and begin writing your post. Let's make it so you can click anywhere inside the body of the editor.

#### Screenshots

The below screenshot was submitted by a user, James, who reported this issue to our support email:

image

Lanny:
## Description

Mainly fixes issues with inline images, namely the inability to move the cursor past an inline image when it's at the end of the document, and adds some visibility into when the image is "selected" (meaning the cursor is on it). But also: Fixes #547 while we're here.

## Migrations

No migs


robin-macpherson (approved):
Great stuff, man! 🙏🏼

PR #551, opened 2021-04-28
Moosend integration

### LINKED ISSUE #418: Mailchimp integration
When users sign up we should add them to mailchimp. Will also require a backfill for existing users.

Lanny:
## Description

Auto-add users to a mailing list on signup.
Closes #418 

## Backfill

Once this code merges, new signups will be added to the "Product Update" mailing list (and the stage variant, in stage), but existing users won't be. To backfill, run the following command:

```
psql $DATABASE_URL -c 'SELECT id, "createdAt", email FROM "User"' --csv > moosend_backfill.csv
```

And upload that CSV to moosend. To prevent any gaps in coverage, this should be done soon _after_ migrations are run and this code hits prod. New users added between deploying and exporting the CSV _should_ be handled on moosend's side.

## Migrations

One migration to add the moosend subscriber ID field to the user model. Should be applied before running code, no custom SQL required. On the whole this is a pretty low risk mig.


packages/web/nexus/utils/email.ts
```diff
@@ -110,3 +126,69 @@ export const sendNewBadgeEmail = ({ user, badgeType }: sendNewBadgeEmailArgs) =>
...
+    console.warn('No `MOOSEND_API_KEY` present, not signing user up for product updates.')
+    return
+  }
+
+  // If user already has an ID, do an update instead (only changes the URL)
+  const url = user.moosendSubscriberId
+    ? `https://api.moosend.com/v3/subscribers/${listId}/update/${user.moosendSubscriberId}.json?apikey=${apiKey}`
+    : `https://api.moosend.com/v3/subscribers/${listId}/subscribe.json?apikey=${apiKey}`
+
+  console.log(url)
```
robin-macpherson, line 152: Is this in here for logging purposes or was it just for dev?

robin-macpherson (approved):
Looks great! Let's have a chat about big-picture merge strategy with #502 💪🏼

packages/web/nexus/utils/email.ts
Lanny: just for dev, removed now.

PR #559, opened 2021-04-30
Added quickly hacked together `grant-free-subscription` script

Lanny:
Lets us give people zero cost subscriptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,22 @@ query feed {
 }
 ```
 
+## Scripts
+
+### greant-free-subscription
+
+Usage:
+
+```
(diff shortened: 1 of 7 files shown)


robin-macpherson (approved):
🦔✨⏰🏀 Lanny with the buzzer beater PR!

robin-macpherson:
@Lanny I'm thinking that since there's some refactoring here maybe we hold on actually merging this so that we can test it a bit more just for sanity sake since we're going live tomorrow, unless you think this is all completely safe / you've tested it thoroughly already then feel free to merge

PR #560, opened 2021-05-01
Upgrade Slate to ^0.62.0

### LINKED ISSUE #537: Get rid of patched slate build
Basically reverse #493 since https://github.com/ianstormtaylor/slate/pull/3698 has merged.

Then retest #479 as it _may_ be fixed. But even if it isn't fixed, we should still push this through since the patched lib was kind of a terrible hack, as long as it doesn't cause a regression.

robin-macpherson:
## Description

**Issue:** closes #537  

Seeks to get us off of our custom/patched slate build and onto the latest Slate release in hopes that this may also address #479 more comprehensively, meaning also for Android.

## Subtasks

- [x] Start by just updating Slate and see what happens/do some testing
- [x] Address any other issues
- [x] Test test test
- [x] Get help from mods & contributors with Android devices to test those as well

## Migrations

FE only, no migs


robin-macpherson:
🏇🏇🏇🏇🏇🏇

Lanny:
Discussed on slack, but for the record: v0.61 doesn't have the commits we need, although master (v0.62.0, unreleased) does. This changes our deps to a custom build of current slate master.

Also includes minor layout changes to the new/edit post pages to make them use the screen real estate better on phones.

Hold to merge until Linda can test on stage.

PR #555, opened 2021-04-29
Translation tools

Lanny:
## Description

Adds a script (`tt.js`) that generates translation templates (CVS files for translators to fill out) and ingests completed or partially completed templates back into the translation files.

Documentation and CI workflow pending

PR #564, opened 2021-05-04
Spanish translations

Lanny:
## Description

Has commits from #555 which should merge first.

Brings in Spanish translations from the google sheet and moves the UI language select widget to be more in line with the profile settings page (see screenshot (and ignore the "yo", heh, translations weren't updated when I took it)).

## Migrations

One mig to update the ui language enum (used for topic translations, as well as on the FE). Should be run before this merges. j-mail will need to be re-deployed. As long as those things happen, this mig is low risk.

## Screenshots
Screen Shot 2021-05-03 at 9 47 08 PM


robin-macpherson (approved):
Awesome stuff!! Love it. Resolved merge conflicts just now and will merge this in 🤩🇪🇸

Issue #566, opened 2021-05-04
Translate stats on profile page


Lanny:
There's a bunch of copy on that page that needs to be translated

PR #569, opened 2021-05-06
Make stats page translatable

Lanny:
## Description

Makes the content on the stats page (except for date stuff, since that's still a bit up in the air) translatable.

Resolves: #566


robin-macpherson (approved):
Woot!

Issue #573, opened 2021-05-12
Keep Email List In-Sync When User Updates Email


robin-macpherson:
#### Description

Currently when a user updates their email address, that change isn't propagated to our email list.
Let's handle that when updating a user's email address to keep things in-sync.

PR #577, opened 2021-05-19
Moosend Backfill II

Lanny:
## Description

**Issue:** #573

The issue isn't quite right, the issue is that we (I) didn't backfill the DB in the right way so if a user who existed before the first backfill updates their email, we'll see they don't have a moosend ID and create a new subscriber for them. This adds a script to backfill moosend subscriber IDs in the DB.

Also has a fix to (hopefully) make excel guess the correct encoding of translation templates.

Also fixes a small visual issue with inline threads having the wrong background color.

Requires the changes from #559 which now merge conflicts with master, so this should be merged **instead of** 559.


robin-macpherson (approved):
Good stuff!

Issue #567, opened 2021-05-04
Add comments (post and thread) to activity stats


Lanny:
Just what it says on the tin

PR #581, opened 2021-05-21
Comments in activity graph

Lanny:
## Description

**Issue:** #567

Includes post comments and thread comments in activity graph on profile/stats page, with toggles to select which sorts of activities are included in the graph

## Migrations

No DB changes, as before the stats query is ad-hoc

## Screenshots

Screen Shot 2021-05-21 at 12 39 20 PM


robin-macpherson (approved):
Great stuff! 🚀

Issue #568, opened 2021-05-05
DEV - Set up Dockerized option for local development


robin-macpherson:
#### Description

@JoshuaBelden brought up that suggested that some contributors may prefer to use Docker for local development and kindly offered to take on the work of setting up the necessary configuration and documentation for this with [This Approach Doc](https://github.com/Journaly/journaly/pull/436)

With that approval, let's do it 🙂

JoshuaBelden:
Hey @robin-macpherson! I can take this one!

robin-macpherson:
Awesome! Thanks again for all the work on this @JoshuaBelden 🙏🏼

Lanny:
Awesome to see this getting under way, in additional the the dev QoL benefits I was thinking this might simplify our automated e2e testing in the future (since that involves standing up a fully functional instance in a CI environment). As mentioned in the other ticket, the build process can but a bit tricky so if you run into any issue with that or anything else let us know, happy to help

robin-macpherson:
That's actually great point @Lanny ! I had been thinking about this just the other day as I keep thinking about our Cypress stuff. Hadn't thought of this but it could make that a lot easier.

JoshuaBelden:
Hey all, just an update. I'm still working through containerization. Without changing a bunch of directory structures to support multiple docker files, I'm trying to set up postgres using strictly docker-compose and a docker file to host the web app. I believe this will work but running into several issues getting the web app up while seeding the database and copying and incorporating different .env files from different sub directories.

All solvable I'm sure, I just need to do some more research on the best approach.

robin-macpherson:
Awesome, thanks for the update @JoshuaBelden ! We used to set up PostgreSQL directly from `docker-compose` back when we were Dockerized before which was just something like:
```
services:
  db:
    image: "postgres:11"
    container_name: "journaly_db"
    ports:
      - "5432:5432"
    volumes:
      - db:/var/lib/postgresql/data
    environment:
      - POSTGRES_USER=${POSTGRES_USER}
      - POSTGRES_DB=${POSTGRES_DB}
      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
    restart: always

volumes:
  db:
```

Good luck and feel free to let us know if we can help!

robin-macpherson:
Also, we always welcome draft PRs just in case you want to put anything up in order to discuss next to your code. But whatever works best for you of course 😊

Lanny:
[...]

PR #582, opened 2021-05-21
Added initial docker files.

JoshuaBelden:
## Description

Adds container support for both the postgres database and the web server.

**Issue:** <#568>

...

https://journaly.com/dashboard/profile/3068

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+node_modules/
+packages/web/node_modules/
+packages/web/.next/
+packages/web/nexus/api.graphql
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,17 @@
-FROM node as build
+FROM node:14 as deps
 
+USER node
 WORKDIR /usr/src/app
-COPY . .
-
-WORKDIR /usr/src/app/packages/web
+COPY packages/web/package* ./
 RUN npm ci
-RUN npm run build
+RUN npm run prisma:generate
+
+FROM deps as build
+USER node
+COPY --chown=node packages/web/ ./
+
+RUN npm run build:nexus
+
+ENV DATABASE_URL=postgres://developer:journaly@journaly_db:5432/journaly_db
 
 EXPOSE 3000
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,5 +13,21 @@ RUN npm run prisma:generate
 RUN npm run build:nexus
 
 ENV DATABASE_URL=postgres://developer:journaly@journaly_db:5432/journaly_db
+ENV NODE_ENV='development'
+ENV APP_ENV='dev'
+ENV APP_SECRET='journaly'
+ENV SITE_DOMAIN='localhost:3000'
+ENV DEPLOYMENT_URL='localhost:3000'
+ENV MAIL_HOST='<mail-host>'
+ENV MAIL_PORT='<port>'
+ENV MAIL_USER='<mail-user>'
+ENV MAIL_PASSWORD="<thats-a-secret>"
+ENV MAIL_SECURE='false'
+ENV NEXT_PUBLIC_STRIPE_PUBLIC_KEY='iLoveMonayyy'
+ENV STRIPE_SECRET_KEY='iLoveMonayyy'
+ENV STRIPE_WEBHOOK_SIGNING_SECRET='iLoveMonayyy'
+ENV STRIPE_MONTHLY_PRICE_ID='iLoveMonayyy'
+ENV STRIPE_QUARTERLY_PRICE_ID='iLoveMonayyy'
+ENV STRIPE_ANNUAL_PRICE_ID='iLoveMonayyy'
 
 EXPOSE 3000
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM node as build
+
+WORKDIR /usr/src/app
+COPY . .
+
+WORKDIR /usr/src/app/packages/web
+RUN npm ci
+RUN npm run build
+
+EXPOSE 3000
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,14 @@ FROM node:14 as deps
 
 USER node
 WORKDIR /usr/src/app
-COPY packages/web/package* ./
+COPY --chown=node packages ./
+
+WORKDIR /usr/src/app/j-db-client
+RUN npm ci
+
+WORKDIR /usr/src/app/web
 RUN npm ci
 RUN npm run prisma:generate
-
-FROM deps as build
-USER node
-COPY --chown=node packages/web/ ./
-
 RUN npm run build:nexus
 
 ENV DATABASE_URL=postgres://developer:journaly@journaly_db:5432/journaly_db
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,8 +16,7 @@ services:
     image: 'journaly:latest'
     depends_on: [journaly_db]
     build: .
-    #command: 'npm -prefix ../j-db-client run reseed-db'
-    command: 'npm run dev'
+    command: bash -c "npm -prefix ../j-db-client run migrate:apply && npm run dev"
     ports: 
       - "3000:3000"
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,25 @@
+version: '3'
+
+services:
+  journaly_db:
+    image: "postgres"
+    environment:
+      POSTGRES_USER: "developer"
+      POSTGRES_PASSWORD: "journaly"
+      POSTGRES_DB: "journaly_db"
+    volumes:
+      - journaly_db:/var/lib/postgresql/data
+    ports:
+      - "5432:5432"
+  
+  journaly:
+    image: 'journaly:latest'
+    depends_on: [journaly_db]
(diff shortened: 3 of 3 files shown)


JoshuaBelden:
@Lanny I seem to be running into a few problems with the initial setup of journaly now. Initial Setup in Step 2 references a `.env.example` file in root but I don't see one. I do see one in `packages/web` but  further instructions during the database setup references the file in root, so I think something is missing.

If I use the one from web, I can move forward, but during the `npm run migrate:up` step, I get the following error:
```
Checking the datasource for potential data loss...
Error: Database error: Error querying the database: db error: ERROR: current transaction is aborted, commands ignored until end of transaction block
   0: migration_core::api::ApplyMigration
           with migration_id="20210312022844-remove-free-n-pro-user-roles"
             at migration-engine/core/src/api.rs:93
```

Do you happen to know what would cause this?

Lanny:
Hey @JoshuaBelden thanks for opening this! I pushed some commits because at least some of the issues you running into were a result of j-db-client being wonky and you need the NPM credentials to fix it, so I went ahead and did it.

For the specific issue with running the migrations, I'm not 100% sure, but we went through a significant prisma upgrade recently (after you started this branch) so updating that seems to fix it. It shouldn't happen now with the merge from master and changes I added to the branch. As for the .env file, that's a good point. The file does live in `packages/web` but I think our docs in the readme are out of date. That file will be necessary for the build though, not sure what the best way to do that is. Maybe if we move the `Dockerfile` that's at the root into `packages/web` that would make it pick it up? Or maybe compose config has a way to name a env file? Those env vars are only relevant to the web package so pulling it up to the root doesn't seem like the way to go.

I made some changes to the docker file, mostly just docker best practices stuff:
- Changed the base image from `node` to `node:14` since that's what we develop against outside of a dockerized env
- Staged build so rebuilds with no dep changes don't have to `ci` again
- `npm run build` runs the next prod build which isn't necessary for development, so changed it to run `prisma:generate` and `build:nexus` and skip the `build:next` part of the build
- Added a docker ignore so we don't have to copy as much into the image

JoshuaBelden:
@Lanny Thanks for the updates, that actually got me quite a ways. The whole docker build process can now launch a postgres database, along with node and the web app. I'm working on getting the database seeded now and running into an issue that I'm not sure if it's my setup or not.

I noticed that the master branch had quite a few changes, include ones that you checked into this branch. I ended up doing a rebase and resolving the conflicts, but that of course requires history changes and a force push. If you still have this branch checked out, you'll have to do a force pull.

I'm trying to add in the step `npm run migrate:up` to the docker-compose process that's mentioned in the README.md, but that script is no longer available. I'm assuming that we should run one of the other migrate steps, such as `apply` or `deploy` but then I get the message:
```Users/joshuabelden/Code/journaly/packages/j-db-client/db-check.js:8
  const matches = url.match(pattern)
                      ^

TypeError: Cannot read property 'match' of undefined
    at url2obj (/Users/joshuabelden/Code/journaly/packages/j-db-client/db-check.js:8:23)
    at Object.<anonymous> (/Users/joshuabelden/Code/journaly/packages/j-db-client/db-check.js:29:22)
    at Module._compile (internal/modules/cjs/loader.js:1157:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1177:10)
    at Module.load (internal/modules/cjs/loader.js:1001:32)
    at Function.Module._load (internal/modules/cjs/loader.js:900:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:74:12)
    at internal/main/run_main_module.js:18:47```
    
This is probably because I'm using the script incorrectly.   
So putting docker aside, what is the correct way to get a newly created postgres database setup with journaly schema?

**Comment by robin-macpherson** [comment]
Hi @JoshuaBelden , a few things I can weigh in on:
- The scripts have changed with the big Prisma upgrade that was done. `migrate:up` is now `migrate:apply` in local and test development environments. I'll make some time to update the docs as soon as I can
- The `seed.ts` script has now been removed because it kept getting out of date and had certain fields missing and ended up being more trouble than help. We'll be retooling things so that we instead seed the database with SQL directly which is probably a better pattern anyway.
- The error you're getting is because the environment variable `DATABASE_URL` hasn't been set which is causing the url match to fail

I hope this helps!

**Comment by robin-macpherson** [comment]
Hi @JoshuaBelden , just wanted to check in on the status of this and if you think you'll be able to continue with it? The work on the Approach Doc started over 7 months ago now and now that we've got the underlying work done for E2E tests with Cypress, we'll need to get this work through soon-ish so that we can leverage Docker as a solution for being able to spin up a Journaly environment & seeded Database every time the tests run in CI. 

Thanks so much for all the amazing work on this so far, I definitely don't want this to seem like anything negative. I just wanted to check in and see where things are since the work has been open for a long time and see if you would like to continue working on it, or if you'd like us to take over this work from here?

I hope you are doing well, friend! 😊

**Comment by JoshuaBelden** [comment]
Hey @robin-macpherson It's really good to hear from you and many apologies for me dropping the ball on this. Honestly, I think I'm going to need help. Getting the database into docker was well in my skill-set, but I'm finding my lack of experience with Prisma and other parts of the stack fairly significant challenges. I've gotten really close with everything containerized but the db bootstrap process isn't quite right. 

I would have reached out earlier but didn't want to be an interrupt. When I wrap some other work this morning, I'm going to get my branch updated and will revisit the issue so I can report current roadblocks in more concrete details.

**Comment by robin-macpherson** [comment]
@JoshuaBelden thanks so much for the quick reply! No worries at all, we'll be totally happy to help out and it's not an interruption! Hopefully this will be a good learning experience for everyone as we get this figured out and you get a better understanding of Prisma and other parts of the stack 💪🏼 If you can report those current issues just whenever you get a chance that would be wonderful - it might take us a few days but we'll for sure get back to you and see how we can figure this out.

I'm really glad you got us started back down this path since now that we're getting to a place where we'll need it for CI stuff, it's great that a lot of the work has already been done thanks to you 😄

**Comment by JoshuaBelden** [comment]
@robin-macpherson For whenever you've got space to put attention on this:

I think I'm really close. After checking out my branch, you can run `docker-compose build` and `docker-compose up`, visit http://localhost:3000 and you should see Journaly up and running. If I inspect the database, I have all of the expected tables created, so I assume that Prisma is doing the correct thing.

My immediate issue is when I try to create a user, I get the following error:
```
Unhandled Runtime Error

Error: secretOrPrivateKey must have a value
Call Stack
ApolloError
node_modules/@apollo/client/errors/index.js (26:0)
next
node_modules/@apollo/client/core/QueryManager.js (88:0)
notifySubscription
node_modules/zen-observable/lib/Observable.js (135:0)
onNotify
node_modules/zen-observable/lib/Observable.js (179:0)
next
node_modules/zen-observable/lib/Observable.js (235:0)
iterateObserversSafely/<
node_modules/@apollo/client/utilities/observables/iteration.js (4:48)
iterateObserversSafely
node_modules/@apollo/client/utilities/observables/iteration.js (4:0)
next
node_modules/@apollo/client/utilities/observables/Concast.js (25:42)
notifySubscription
node_modules/zen-observable/lib/Observable.js (135:0)
```

So with that, it would be helpful if someone on your end with a clean checkout of my branch could run docker-compose and just verify that the containerization process is working on any machine other than mine. The error above appears to be related to missing configuration, which brings up a bigger struggle of mine. There was a bit of confusion on how exactly to work with the .env files and what directory they're supposed to go in. At the time, the readme was slightly off and not sure what's accurate. My hope is that I've simply missed a configuration value somewhere.

robin-macpherson:
Thank you for the really helpful report @JoshuaBelden ! I'll get to looking at this as soon as finishing up a few things I've got on the go at the moment. Will keep you updated!

Lanny:
Hi @JoshuaBelden sorry for the delay getting back to you. The error you're seeing is because `APP_SECRET` isn't getting set. That env var is used for JWT signing (which is used for auth). In theory you can set it to be any string, we typically use the value of `journaly` for dev but it doesn't really matter, as long as it's consistent between builds (so that sessions aren't invalidated when you rebuild).

Setting it in the docker file itself is probably fine, same as how the `DATABASE_URL` env var gets set

JoshuaBelden:
@robin-macpherson @Lanny Looks like this worked. I ended up copying the env variables from the .env.example file into the Dockerfile. I'm able to get the site up and running and can now post articles. If this seems like the right approach, we'll probably want to update docs. I'm still not sure what the expected use of the .env.example files are, or where they're required.

Lanny:
Cool, this works for me for now.

The `.env.example` file is there as a template for creating a `.env` file for running the code un-dockerized locally. If you copy a .env file into the docker image the webserver will pick up on it, but then your env file is burned into the image and it kinda defeats the purpose. We'll probably look at making these vars defined in the Dockerfile configurable by something at runtime (need to look at what Docker options for that are) but if we're in a state where someone with only docker can get the project up and running with just `docker-compose up` I think that's a win and we can worry about runtime config when we start looking at using docker for e2e testing

robin-macpherson (approved):
Fantastic work everyone! 🥳 🙌🏼 Thanks so much @JoshuaBelden for driving this & @Lanny for the support.

This now opens up work on #609 as well!

JoshuaBelden:
Thanks everyone. Sorry that took so long. Longest PR ever!

PR #585, opened 2021-05-28
Add More Default Images

### LINKED ISSUE #557: Add a few additional "default" images for posts and select one at random
#### Perceived Problem

- We currently have one default image that gets used if a user does not upload their own post image
- We've now all seen that typewriter a million times 

- Let's upload 4 more defaults and write some simple logic to select one of them at random as the default image
- Since we show the default image in the post preview, it's probably best to select the image when loading the "new post" page which would also allow the user to refresh if they wanted a different one

#### Future iterations

- Once we build the Unsplash integration and UI, this could be included and even expanded within that UI, very similar to how Notion does it

robin-macpherson:
## Description

**Issue:** closes #557 

- Adds a few new default images so that we have a total of five
- Images are randomly selected at the time of loading the **New Post** page so that if a user specifically wants a different one, they can simply refresh the page
- All images are uploaded to S3, including the original default image that was sitting in our repo in two locations (3.5MB x 2) so those have also been removed
- Improves things so that images using default still have that field populated rather than 

## Subtasks

- [x] Upload the images to S3
- [x] Write a wee function to randomly select and get that working
- [x] Light refactor so that we actually save the default post image vs. loading it in code each time
  - [x] Discuss this first

## Migrations

_**As currently written, will require a backfill for existing posts that didn't have a default image**_


packages/web/graphql/fragments.graphql
```diff
@@ -161,7 +160,7 @@ fragment PostCardFragment on Post {
   publishedLanguageLevel
   commentCount
   status
-  images {
+  headlineImage {
     smallSize
```
Lanny, line 164: let's add in the ID for cache updating

packages/web/resolvers/post.ts
```diff
@@ -548,28 +538,23 @@ const PostMutations = extendType({
...
-
-          if (headlineImage) {
-            await ctx.db.image.deleteMany({
-              where: {
-                postId: args.postId,
-                imageRole: 'HEADLINE',
-              },
-            })
+        if (args.headlineImage) {
+          await ctx.db.image.deleteMany({
```
Lanny, line 542: Think this can become a `delete` rather than `deleteMany`. Not particularly important though.

Lanny: Actually on second though, since the relation is on the post side, I think we can just leave the image as-is. That way if we ever want to do a sweep to remove unused images from S3 we'll still have the record. Plus it saves us a query here.

Lanny (approved):

packages/j-mail/package.json
```diff
@@ -32,7 +32,7 @@
     "typescript": "3.8.3"
   },
   "dependencies": {
-    "@journaly/j-db-client": "^11.0.0",
+    "@journaly/j-db-client": "^12.0.0",
```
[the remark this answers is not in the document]
robin-macpherson: Great catch!

packages/web/components/Dashboard/PostCard/PostCard.tsx
```diff
@@ -46,15 +46,13 @@ const PostCard: React.FC<Props> = ({
...
-  const displayImage = images.length ? images[0].smallSize : '/images/samples/sample-post-img.jpg'
-  const imageAlt = images.length === 0 ? 'Typewriter on an old wooden desk' : ''
   const postCardStyles = classNames('post-card-container', { stacked })
 
   return (
     <>
       <Link href={'/post/[id]'} as={`/post/${id}`}>
         <a className={postCardStyles}>
-          <img className="post-image" src={displayImage} alt={imageAlt} />
+          <img className="post-image" src={headlineImage.smallSize} alt='' />
```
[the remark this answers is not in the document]
robin-macpherson: Yep good point! Previously that variable was just an empty string unless it was the default image so I was more just cleaning things up but definitely worth making this better while we're here!

packages/web/graphql/fragments.graphql
robin-macpherson: Good call! 👍🏼

packages/web/resolvers/post.ts
robin-macpherson: Cool sounds good! So just to be clear, you meant let's not delete the image at all here, right? That's the change I've made but just wanna make sure we're on the same page

PR #587, opened 2021-05-30
[j-mail] Add prisma as a direct dep and upgrade typescript to 4.x

Lanny:
Upgrade typescript for j-mail and ensure prisma:generate runs before a deploy.


robin-macpherson (approved):
Awesome! Thanks for adding that prisma generate guard on the deploy scripts 👍🏼

PR #588, opened 2021-05-30
Fix J-Editor Toolbar Transparency Bug

### LINKED ISSUE #586: Journaly Editor Toolbar is un-clickable when floating
#### Description

The Journaly Editor toolbar now seems to have the wrong z-index when floating and therefore appears transparent and when you try to click its buttons, you end up clicking the text behind it. Therefore, it's impossible to use the toolbar on elements of your post after it gets beyond a length that fits on your viewport.

#### Screenshot

Screenshot 2021-05-29 at 08 56 01

#### Repro Steps

1. Write a post, enter lots of new lines until the post no longer fits on viewport and scroll down to activate the floating toolbar

robin-macpherson:
## Description

**Issue:** fixes #586 

The toolbar ended up with appearing as transparent when floating above elements and was actually behind them making it also unclickable.

This PR simply sets the z-index to `1` which appears to be sufficient.

## Subtasks

- [x] Make the fix

## Migrations

No migs

## Screenshots

(See issue for original problem)
image

### DIFF
--- a/packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
+++ b/packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
@@ -161,6 +161,7 @@ const Toolbar = ({ allowInlineImages }: ToolbarProps) => {
           justify-content: center;
           padding: 15px 0;
           border-bottom: 2px solid #eee;
+          z-index: 1;
         }
 
         .is-fixed .editor-toolbar {


robin-macpherson:
@Lanny going to merge this in since it's a very minor change but currently the bug makes the toolbar totally unusable if your post is a decent length, but would you mind having a look as a sanity check whenever you get a chance?

PR #595, opened 2021-06-02
Ensure no extra fields show up when saving a post with an existing image

Lanny:
Presence of `id` and `__typename` were throwing gql for a loop when saving a post without changing the headline image.


robin-macpherson (approved):
🚑💨 ... 🦔💨🍃

PR #596, opened 2021-06-02
Set max vert size on inline images

Lanny:
## Description

On laptop resolutions (wide, limited vertical resolution) images could end up occupying more the the full view port, this PR limits the vertical size of an image to 75% of the VP height. 

@robin-macpherson you mentioned that you thought it was going to be more complicated than this, but just a simple limit on vertical size seems to work. Some screenshots below. Let me know if any of these seem off.

Also removes layout padding on posts which I think gives post bodies some needed additional real estate

## Screenshots

Posts with layout padding (existing):

Screen Shot 2021-06-02 at 1 46 33 PM

Without layout padding (proposed):

Screen Shot 2021-06-02 at 1 46 05 PM

Vertical size limiting:

Screen Shot 2021-06-02 at 1 59 00 PM
Screen Shot 2021-06-02 at 1 59 03 PM
Screen Shot 2021-06-02 at 1 59 09 PM
Screen Shot 2021-06-02 at 1 59 14 PM

### DIFF
--- a/packages/web/components/PostBodyStyles/PostBodyStyles.tsx
+++ b/packages/web/components/PostBodyStyles/PostBodyStyles.tsx
@@ -42,6 +42,7 @@ const PostBodyStyles: React.FC<Props> = ({ parentClassName }: Props) => {
 
       .${parentClassName} img {
         max-width: 100%;
+        max-height: 75vh;
         margin: 0 auto;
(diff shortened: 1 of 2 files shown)


robin-macpherson (approved):
Sweet!

PR #598, opened 2021-06-05
FEATURE: Clapping for posts! 👏🏼

### LINKED ISSUE #480: Build Post Liking
#### Background

There are two different concepts at Journaly that are often a single concept of "liking something" on many apps:

1. the concept of "thanks" which we associate with inline feedback comments where you can say thank you to someone for a piece of feedback
2. the concept of "liking" a post in general

We want to be able to incentivize and reward helping fellow community members, and we also want a way that high-quality/helpful feedback can be identified over low-quality/less helpful feedback. This concept of being able to say "thanks" for a feedback comment created a way to do this, also allowing us to aggregate data about things like how many "thanks" a user has received as one measure of how much help they have contributed to the community.

This differentiation allows for "liking" a post to be a little more generic. It's a way of indicating you like something, and also the popularity of a post without this impacting the previous concept of thanks and a user's related statistics.

#### Perceived Problem

- Currently, the functionality for liking a post is not yet built
- Because we display the lake count in the postcards in the feed, and also a users profile, It can be confusing for some users who might think this is supposed to be related to the number of thanks that have been given/received inside of a post
- Currently the same icon is used for thanks and also the lake account, which can further lead to confusion

- I would like to find two separate icons for thanking and liking To help us more strongly differentiate these two concepts with our visual design language
- I've struggled quite a bit figuring out what icons would be best for this.
  - I thought about prayer hands which I quite enjoy using myself colloquially, but these have a religious connotation that make them potentially inappropriate for a global and inclusive platform
  - So far for thanking, the heart seems like a reasonable choice
  - For liking, I wanted to pick something else that is not the thumbs up. This is another gesture which can mean different things in different cultures, and also feels very Facebook-y 😒
  - As a short term solution, I thought about using a star for the liking concept, although this has been used a lot for the idea of a favorite on other platforms

robin-macpherson:
## Description

**Issue:** closes #480 

FINALLY builds one of (maybe literally?) our last core/essential features, but also involves some nice redesign of the overall concept.

### UX / Concept Redesign Notes

- Originally, we thought of this as "liking" a post but we had some trouble thinking of things like an icon we were totally happy with that and also how to clearly differentiate a "like" from a "thanks" given to comments.
- I finally landed on what feels like a great solution. In the context of Journaly, what really is the point of a "like"? To me, something that fits this space far better is the idea of encouraging the author. Sort of a "well done!" for their efforts. This is how I landed on the idea of switching our "likes" for "claps", which is something that Medium also does!

---

- Includes optimistic UI implementation 
- Includes migration to replace/clean up `PostLike` table with `PostClap` but also includes adding claps to our notifications and daily digest email

## Subtasks

- [x] Schema changes & migs
- [x] Resolvers
- [x] New GraphQL mutations/updates
- [x] Choose new icon for clappies 👏🏼
- [x] frontend visual updates
- [x] Wire up mutations
- [x] Implement optimistic UI
- [x] Handle own user case + other post actions
- [x] Test visually all device sizes

### Deployment
- [x] 🚨 Deploy migs to stage
- [x] 🚨 Deploy code to stage
- [x] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

Contains migs to replace the `PostLike` table with `PostClap` as well as adding claps to the pending notifications table and related enum.

- Should be applied before deploying the code
- These should be safe as the existing table is empty in prod
- No backfill required


packages/web/components/Dashboard/Post/Post.tsx
```diff
@@ -289,6 +294,80 @@ const Post = ({ post, currentUser, refetch }: IPostProps) => {
...
+        __typename: 'Mutation',
+        createPostClap: {
+          __typename: 'PostClap',
+          id: generateNegativeRandomNumber(),
+          author: {
+            __typename: 'User',
+            handle: currentUser!.handle,
+            name: currentUser?.name,
+            profileImage: currentUser?.profileImage,
+            id: currentUser!.id,
```
Lanny, line 326: The non-null assertion right next to an optional chain on the same object looks strange. If you're going to assert `currentUser` isn't null, it means accepting that the type system can not promise you this won't throw at runtime, so optional chaining is pointless (e.g. if `currentUser` is null or undefined, then line 323 will throw, the optional chain on 324 is irrelevant since the only case where it makes a difference is unreachable).

If you know for sure that current user can't be null here, then you can use the non-null assertion in all four cases

Lanny (approved):
Looks good! Just one typescript related nit pick, but I'll go ahead and approve since it should be a real quick fix.

packages/web/components/Dashboard/Post/Post.tsx
robin-macpherson: Sounds good! i went ahead and added `|| !currentUser` to the disabled attribute on the button as well!

PR #603, opened 2021-06-13
Old post feedback badge

### LINKED ISSUE #433: Badge: Leaving feedback on posts of a certain age or greater
Add a badge for people who go to the trouble of leaving feedback on posts that are "old", maybe for feedback given on posts over a week old.

Lanny:
## Description

Closes #433

Adds a badge ("Necromancer", which is terminology borrowed from SO. I couldn't think some something concise and better) that is assigned when leaving feedback on a post authored by someone else that's over a week old.

Also adds an easter egg badge that's assigned for saving a post without any edits.

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

One mig, low risk as all it does is expand an enum

### DIFF
--- a/packages/j-db-client/package.json
+++ b/packages/j-db-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@journaly/j-db-client",
-  "version": "13.3.0",
+  "version": "13.4.0",
   "description": "Journaly's internal database client.",
   "main": "dist/index",
   "scripts": {
--- a/packages/j-db-client/prisma/migrations/20210612192514_add_odradek_and_necromancer_badges/migration.sql
+++ b/packages/j-db-client/prisma/migrations/20210612192514_add_odradek_and_necromancer_badges/migration.sql
@@ -1,0 +1,10 @@
+-- AlterEnum
+-- This migration adds more than one value to an enum.
+-- With PostgreSQL versions 11 and earlier, this is not possible
+-- in a single migration. This can be worked around by creating
+-- multiple migrations, each migration adding only one value to
+-- the enum.
+
+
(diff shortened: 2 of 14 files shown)


robin-macpherson:
Awesome stuff @Lanny ! Just two small discussion points:
- I'm not so sure about the names 🧐 If I saw these names I personally would have absolutely no clue what either of them meant 😅 I feel like if it's fine if the name is something that I'm not familiar with but when I see the explanation I'm like "Ooooh I get it!", but with both of these I see the explanation and then still feel the need to go and google the terms, and even then it feels like a stretch or not totally clear. Especially thinking about how so many of our users won't be native English speakers. WDYT?
- Think it's worth trying to do a backfill on this? I guess that'd be a pretty interesting one to write haha

Lanny:
> I'm not so sure about the names

Totally fair. Do you have suggestions for names? "Odradek" comes from a short story by Kafka which is admittedly a bit of an obscure reference, but since it's an easter egg it didn't bother me. I'd be all for a better name for giving feedback on an old post but really couldn't think of something brief that worked better (the etymology is from like forums or BBSs where posting in an old thread "brings it back from the dead" by returning it to the list of recently active thread, sometimes referred to as "necrobumping", but has since spread as evidenced by SO)

> Think it's worth trying to do a backfill on this

I think it would be OK to not backfill, since it's still easy for people to earn it who may have already done the required action. Also the backfill would involve some fancy date arithmetic that I don't think we want to dive into

robin-macpherson (approved):

PR #614, opened 2021-07-13
added placeholder if user has no languages in settings

### LINKED ISSUE #451: Make it clear a user needs to add languages if dropdown is empty on new post
#### Perceived Problem

- Some users see an empty dropdown when writing a new post due to having not added language info yet in their settings and think this is a bug

- Let's change the placeholder if it's empty to help them understand why it is empty and what to do next

#### Screenshots

image

dom-waterson:
## Description

Make it clear a user needs to add languages if dropdown is empty on new post

**Issue:** closes #451

## Subtasks

- [x] Added new placeholder if user has no languages setup in settings
- [x] Added translations

## Screenshots
Screenshot 2021-07-13 at 19 21 33

### DIFF
--- a/packages/web/components/PostEditor/PostEditor.tsx
+++ b/packages/web/components/PostEditor/PostEditor.tsx
@@ -39,7 +39,7 @@ type InputPostData = BasePostData & {
 }
 
 type PostEditorProps = {
-  currentUser: UserType,
+  currentUser: UserType
   autosaveKey: string
   dataRef: React.MutableRefObject<OutputPostData | undefined>
   initialData: InputPostData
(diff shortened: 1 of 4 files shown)


robin-macpherson (changes requested):
Thanks a lot for the PR, @dom-waterson ! In general, we ask that you keep your change set to only include your changes to the code, and avoid things like purely auto-formatting changes (for example your PR has some formatting changes to code that isn't relevant to this work).

Also, we have a mechanism for community sourcing translations to make sure everything is totally natural in each language. Are you a native German or Spanish speaker? If so, we can totally keep your translation, but otherwise let's remove it so that our translation tooling can pick it up 😊

Otherwise, looks good! 👍🏼

dom-waterson:
@robin-macpherson Thanks for checking on this. Removed unrelated formatting and translations. Currently learning Spanish so will leave them out from now on :)

robin-macpherson (approved):
Amazing, thanks so much for this @dom-waterson and I apologise again for the delay in getting it through! Things have been a bit crazy. Very happy to have this simple but really important improvement done! 🙏🏼

PR #622, opened 2021-07-29
Bump our patched slate version.

Lanny:
## Description

**Issue:** Maybe #479

Bumps our patched slate version to disable android specific editor, which for some reason makes IME on android actually work. Opened an issue against slate here: https://github.com/ianstormtaylor/slate/issues/4400

Newly patched version _may_ resolve the issues android users have been experiencing with repeated characters when using IME, or failure to compose multi-keystroke characters at all (latter was something I experienced but not sure if it was ever reported).

@robin-macpherson can you reach out to some of the folks that reported issues on Android to see if they can test in the preview environment?

Let's not merge this until we get feedback. It pulls in hundreds of commits from slate master, some of which I expect are pre-release, so it represents a decent amount of risk of breakage.

## Migrations

No migs, FE only.


robin-macpherson:
Thanks for this @Lanny ! Unfortunately, first Android test results 😢
image

robin-macpherson (approved):
Ahhhh I hope this fixes issues for at least some folks! 🤞🏼✨✨✨

Issue #512, opened 2021-03-02
Display "thanks" count on a user's own comments


robin-macpherson:
#### Perceived Problem

- Currently, a user cannot actually see a total count for how many thanks one of their own comments has received
- One of the big reasons for having this thanks feature is to make users feel rewarded for their efforts and time to help the community. Without this feedback on how useful their post has been, we may lose this benefit or users may even become demotivated

#### Ideas / Proposed Solution(s)

- Let's display the thanks count as we do for everyone else
- Let's disable thanking yourself

PR #624, opened 2021-07-31
Display number of likes in a user's own comments

meredithcat:
## Description

**Issue:** #512

I added a non-clickable like icon to the user's own inline comments, if the comment has at least one like.

I thought it looked too busy to put the icon side-by-side with the edit/delete icons, so I placed it above. It was a bit tricky to get the CSS right, and there might be a better way that doesn't result in so many new CSS classes.

Journaly.com profile url (include if you'd like the "code contributor" badge): https://journaly.com/dashboard/profile/8654

## Subtasks

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Screenshots

Short comment with likes:
Screen Shot 2021-07-30 at 9 41 41 PM

Long comment with likes:
Screen Shot 2021-07-30 at 9 41 25 PM

A comment without any likes appears the same as before:
Screen Shot 2021-07-30 at 9 41 50 PM

### DIFF
--- a/packages/web/components/InlineFeedbackPopover/Comment.tsx
+++ b/packages/web/components/InlineFeedbackPopover/Comment.tsx
@@ -186,14 +186,15 @@ const Comment = ({ comment, canEdit, onUpdateComment, currentUser }: CommentProp
       </div>
       {canEdit && !isEditMode && (
         <div className="edit-thanks-block">
-          {numThanks > 0 && (
-            <div className="thanks-block">
-              <span>
-                <LikeIcon filled={true} />
-              </span>
-              <span className="thanks-count">{numThanks}</span>
-            </div>
-          )}
+          <div className="thanks-block">
+            <span>
+              <LikeIcon 
+                filled={numThanks > 0}
+                title={t('numUsersGaveThanks', {numThanks})}
+              />
+            </span>
+            <span className="thanks-count">{numThanks}</span>
+          </div>
           <div className="edit-block">
             <span
               role="button"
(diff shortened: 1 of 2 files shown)


robin-macpherson (changes requested):
Great work! 😄

Still looking through the code and testing but as some early feedback, I like the approach of placing the icon above as you did 👏🏼 One change I'd like to request is for us to simply display the thanks count always, even if it's 0. I can totally see the reason for not showing it but I'd like to keep things as straight-forward and visually consistent as possible. For example we do show 0 likes when not in edit mode and I just want to make sure it's perfectly clear to the user.

Can't wait to get this one through! 🏁 . . . 🦔💨🍃

meredithcat:
> One change I'd like to request is for us to simply display the thanks count always, even if it's 0. I can totally see the reason for not showing it but I'd like to keep things as straight-forward and visually consistent as possible. For example we do show 0 likes when not in edit mode and I just want to make sure it's perfectly clear to the user.

Done, plus I added a tooltip to explain the purpose of the icon. I added the tooltip text to the i18n `en` file, but I don't speak German or Spanish well enough to translate! How should I proceed?

robin-macpherson:
> Done, plus I added a tooltip to explain the purpose of the icon. I added the tooltip text to the i18n en file, but I don't speak German or Spanish well enough to translate! How should I proceed?

Thanks for adding the tooltip 🙂

We actually have some exciting translation tools that we built now which can be found [here](http://translations-website.s3-website.us-east-2.amazonaws.com/), so unless you speak a language natively or are really confident your translation is correct and natural, it's best to actually leave the translations out so that our translation tools pick those up as missing and community translators can see them flagged as needing translation. We've been testing them out carefully with just a couple of people and languages but will soon roll the translation tools out more publicly 😊

robin-macpherson (approved):
Looks great! Thanks again so much for tackling this and for being very thorough 🎈🦔

Issue #606, opened 2021-06-21
👬🏻 Following Experience


robin-macpherson:
#### Perceived Problem

- Currently, our whole UX (and UI) around following or being followed by users is very rudimentary and I think it is easy to feel quite lost within it
- It does provide the utility of being able to filter the feed to posts from Journalers you follow, but that's it really

#### Ideas / Proposed Solution(s)

- Let's allow users to see a list of Journalers they follow and are being followed by (#429) 
- Display following/follower counts?
- Add the ability to follow a user directly from their profile

Issue #583, opened 2021-05-25
Profile Page Improvements


robin-macpherson:
#### Description

The profile page looks really nice but certainly has some places where it could be improved

#### Ideas

- There is a visual bug where on wide screens (wider than a 15-inch macbook pro screen) the height of the profile is too short that has been around for a while
- Adding a button to follow this user on their profile would be a huge UX improvement to the general flow of following people
- We really need to take care of #453 and add infinite scrolling or some sort of pagination for the user's posts on their profile, as some users are now accumulating a lot of posts (which is awesomeeee 🙌🏼). I think this is a good case for infinite scroll
- With #574 , we'll have some nice visual improvements to the badges section
- Possibly rethinking where we put a user's public-facing stats when #576 is completed
- Rethink the language used in the profile around "speaking" vs. "learning". Maybe simplifying this to simply mirror the "native, advanced, intermediate, beginner" levels that we use so that it is consistent

meredithcat:
Would the user's likes fit in this issue as well? I noticed that we are currently showing hard-coded likes of `['languages', 'journaling']`.

PR #626, opened 2021-07-31
Add follow button to profile page

meredithcat:
## Description

**Issue:** #583, #606

Add a follow button to the profile page, unless viewing own profile.

The code here is a bit repetitive as I copied it from the `PostAuthorCard` component, so we could abstract it into a `FollowButton` component.

Journaly.com profile url (include if you'd like the "code contributor" badge):

## Subtasks

- [ ] ...

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Screenshots

Without badges:

Screen Shot 2021-07-30 at 10 24 38 PM

With badges:

Screen Shot 2021-07-30 at 10 25 10 PM


robin-macpherson (changes requested):
Same comments as @Lanny regarding the sizing, would definitely be a great opportunity to make those improvements. One small design change I'd like to request is to use `ButtonVariant.Link`. Definitely a small thing and is purely aesthetic, but I've tried to design the profile page to be really nice and sleek and I think the big blue button looks a little out of place with the rest of the design, but I think a nice Journaly Blue link might look quite nice! Also keeps it visually consistent with the Follow button in the `PostAuthorCard`. If it feels like it needs a bit of extra visibility maybe we could try bumping up the font weight a wee bit. Let me know what you think! 😊

Other than that awesome work!

meredithcat:
Hmm, yeah @Lanny that is an important issue to consider. I'd lean towards just sticking in a scrollbar for the left-hand side, since like you said, it doesn't happen often (only for a few specific profiles & screen sizes). But maybe it's worth thinking of some other ideas first since it seems like we might be making more additions to the profile page.

@robin-macpherson I did slightly prefer the look of the button to the link, but I'll defer to your judgment and make that update!

meredithcat:
Here is another layout idea, this time with the `ButtonVariant.Link` type and a slightly larger font size.

I still don't love either the `ButtonVariant.Link` or the `ButtonVariant.Primary` options (although, I still prefer the latter as it draws the eye a bit more and is less likely to get lost with all of the other text). Any chance we could get a more dark theme friendly button type to fit in more with the background?

Screen Shot 2021-08-02 at 7 32 23 PM

robin-macpherson:
@meredithcat thanks for giving those things a try and presenting other options. For now I'd still like to go with the `ButtonVariant.Link` option as you showed in your alternate layout screenshot, but in the original position below the user's avatar.

In terms of my reasoning: I see what you're saying about the button being more eye-catching, but my design philosophy with the profile page was for things to be very elegant and subtle but still easy to find, since there is a decent amount of information on the page. If everything is designed to be as eye-catching as possible then nothing ends up standing out. In this case, I don't think the "Follow" button is the most important thing that people need to see immediately, as long as it's still perfectly visible and easy to find. We can definitely revisit in the future and look at other `ButtonVariant`'s but in trying to keep the design system as minimal as possible, adding new variants, colors, etc. is something that if possible I want to always think through and make sure there's a broad use case for vs. making a new variant for a specific situation.

Still love and welcome the discussion and different options! But I think for the moment, that fits the design style of the page well and still gets the job done, adding lots of value since it's definitely intuitive that you'd be able to follow a Journaler from their profile! 🎉

robin-macpherson:
Just to add one more thought, I do like the way you alleviated space constraints with your alternative option, so if we do still run into issues like the one shown in this screenshot, I'd say we could totally go with your alternate design, even as a really nice way to still deliver this important feature without having to get too stuck in the weeds at this moment 😊

Screenshot 2021-08-04 at 09 53 59

meredithcat:
Just an update - I don't think I will have the bandwidth to finish up this change since it looks like it may need some more UI tweaks. So, if someone wants to take it over, please go ahead as I think it is close to done 😬

robin-macpherson:
@Lanny just made a few small adjustments and tightened up spacing a tiny bit but it doesn't help the issue too much. I tried putting an `overflow-x: hidden` on `.profile-card` to enable scrolling within `ProfileCard` but that then causes issues with the background image sizing, which I tried various things to address but kept running into more issues. Any thoughts?

Lanny (approved):

Lanny:
Think I have a solution to the "profile card always scrolls". The container needs to be a fixed height, but the image needs to be the size of the content (which may be larger than the container and scroll), so I added another container that holds both the content and the image. Image matches the size of the inner container, while the outer container scrolls. Works for me at the screen sizes I tried. Everything else looks good to me.

Give it a whirr and if it works for you I think we're good to go

robin-macpherson:
@Lanny WOOHOOOO nice work, looks so much better! Good thought with adding that extra container. Man so happy to get these changes in 💪🏼 Thanks all 🚀

robin-macpherson (approved):

PR #632, opened 2021-08-02
Show a tooltip with the specific users who clapped for a post

### LINKED ISSUE #631: Show tooltip on clap to see which users clapped
#### Perceived Problem

Currently, there is a way to see how many claps a particular post has, but no way to see who gave a clap.

Screen Shot 2021-08-01 at 4 36 25 PM

We should show a tooltip on the clap icon to show the list of users who gave a clap. If a large number of users clapped, we could truncate the list and show something like: `Jon Snow, Arya Stark, and 4 others clapped.`

For posts where the user clapped, it would be better to show something like: `You, Jon Snow, and 5 others clapped.`

meredithcat:
## Description

**Issue:** closes #631 

One potential issue with my code is that it was quite messy to put the type of the `claps` array in `helpers.ts`. I couldn't figure out a better way (the type checker didn't like the `PostClap` type), but if there is one, please let me know!

## Subtasks

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Screenshots

Screen Shot 2021-08-01 at 6 26 04 PM

### DIFF
--- a/packages/web/public/static/locales/en/post.json
+++ b/packages/web/public/static/locales/en/post.json
@@ -42,7 +42,7 @@
   "websitePatternError": "Invalid website url. e.g. domain.com or sub.domain.com",
   "claps": {
     "currentUserPronoun": "You",
-    "noUsersClapped": "0 users clapped for this post.",
+    "noUsersClapped": "No users clapped for this post... yet!",
     "oneUserClapped": "{{name1}} clapped for this post.",
     "twoUsersClapped": "{{name1}} and {{name2}} clapped for this post.",
     "manyUsersClapped": "{{name1}}, {{name2}}, and {{numOthers}} others clapped for this post."
--- a/packages/web/public/static/locales/en/post.json
+++ b/packages/web/public/static/locales/en/post.json
@@ -39,5 +39,12 @@
     "body": "Are you sure you want to cancel editing this post? Unsaved progress will be lost."
   },
   "enterUrlPrompt": "Enter the URL of the link:",
-  "websitePatternError": "Invalid website url. e.g. domain.com or sub.domain.com"
+  "websitePatternError": "Invalid website url. e.g. domain.com or sub.domain.com",
+  "claps": {
+    "currentUserPronoun": "You",
+    "noUsersClapped": "0 users clapped for this post.",
+    "oneUserClapped": "{{name1}} clapped for this post.",
+    "twoUsersClapped": "{{name1}} and {{name2}} clapped for this post.",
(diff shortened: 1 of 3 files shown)


robin-macpherson:
Hi @meredithcat ! It's a wonderful idea to include this nice and simple way for people to see a list of the users who clapped for them! Maybe in the future we can implement it as a simple modal so that we can handle longer lists and so that people can click on a user who clapped for them to go to their profile and things like that. This could be nice when we have direct messaging, too. But anyway, that's just me rambling about future ideas – for now I think this is a fantastic and simple solution that gets the job done 🎉

I believe the troubles you were having with the type-checker and using the `PostClap` is that you need to create a GraphQL fragment that contains just the fields you expect and then import that. When you import the generated type directly it has absolutely all the fields on it which the GraphQL calls aren't expecting. For a reference, you can take a look at what we do with `ProfileUserFragmentFragment` in `ProfileCard.tsx` (the naming ends up being a little goofy). I think this should solve your issue and simplify things but let me know if you still have trouble!

robin-macpherson:
By the way, if you use the keyword "closes" in front of the issue, GitHub will automatically close the issue when your PR gets merged. "fixes" also does the same thing so we use that with bugs. I updated your description accordingly. Just a tiny thing of course but a nifty GH feature that saves a few clicks closing issues! 😄

packages/web/public/static/locales/en/post.json
```diff
@@ -39,5 +39,12 @@
     "body": "Are you sure you want to cancel editing this post? Unsaved progress will be lost."
   },
   "enterUrlPrompt": "Enter the URL of the link:",
-  "websitePatternError": "Invalid website url. e.g. domain.com or sub.domain.com"
+  "websitePatternError": "Invalid website url. e.g. domain.com or sub.domain.com",
+  "claps": {
+    "currentUserPronoun": "You",
+    "noUsersClapped": "0 users clapped for this post.",
```
robin-macpherson, line 45: I think this could be a nice opportunity to further our fun and playful tone. Perhaps something like at least adding "...yet!" on the end, but I wonder if there's something even more creative 🧐 I'll have a think an happy to get your ideas if you have any! This is of course a super minor thing 😆

meredithcat: Done

meredithcat:
I have incorporated your suggested changes! Should be ready now

robin-macpherson (approved):
Looks fantastic! Great work 😄🚀

Issue #627, opened 2021-07-31
Allow users to add/view user interests on settings/profile


meredithcat:
#### Perceived Problem

Currently, users cannot add or view their interests. Instead, we show hardcoded values of `languages, journaling`.

#### Ideas / Proposed Solution(s)

Adding interests to the user profile flow would require the following:

- Add interests to the prisma schema (looks like this is already done, with the `userInterests` field)
- Update the resolvers to send/receive the user interests in the profile/settings commands
- Add a drop-down for user interests to the settings page, similar to the topic filter in the posts dashboard
- Display user interests on the profile page
- (Optional) Add a "My Interests" filter on the posts dashboard, similar to the "My Languages" filter
  - Would need to decide whether to place it below the topic picker, or below the language picker with the rest of the filters - might become too bloated though if we have too many filters there. Looks like this would fit well with the proposed update in #617.

meredithcat:
I don't think I can assign an issue to myself, but I can take this one on if it's not too urgent. Might be a good one to see how the whole tech stack fits together. @robin-macpherson

robin-macpherson:
@meredithcat this would definitely be a great thing to get done as it's one of the very last core things that doesn't work yet. I should say this one is a little more involved than it first looks due to some updates and improvements we've made to the schema design over time since first modeling the notion of `UserInterests`. You're still welcome to take it on, which would be amazing 💪🏼  But I'd just want to give you some helpful context first and I'd like to potentially propose a few schema changes before we get started.

I was just writing out some context and technical design proposals but it's almost bedtime here and I'm hoping to squeeze in a little Japanese study before bed. If it's okay, I'll come back to you tomorrow with more details 😊

Issue #617, opened 2021-07-28
"Advanced Filters" UI Updates To My Feed page


robin-macpherson:
#### Perceived Problem

- Currently, the various filters and filter actions take up quite a bit of screen real estate and we have more on the way
- This is especially true on mobile devices where the current filters take up about 65% of the vertical screen space when first loading the My Feed page

#### Ideas / Proposed Solution(s)

- The "languages" filter is the most important and widely used, so let's keep that one always showing by default, but let's hide everything else within a simple "advanced filters" menu that can be expanded/collapsed easily
- We may also want to revamp things a little and make some of the filter actions check boxes that can be checked/unchecked
  - followed journalers
  - needs feedback (coming soon)
  - posts I've commented on (coming soon)
  - my languages (?)
  - my interests

PR #633, opened 2021-08-04
Add user interests to settings & profile page

meredithcat:
## Description

**Issue:** #627 

One potential issue is that I had to add the $uiLanguage variable to a bunch of other user queries, so that the user interests could be added to the queries. If this is not desired, then we could create a new user fragment that contains the interests, specifically for the pages where it is needed, but that seemed like overkill.

Still to do is adding a custom filter for "My Interests" to the feed page, so let me know if I should include that in this PR or a future one.

## Subtasks

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Screenshots

Screen Shot 2021-08-03 at 7 31 08 PM
Screen Shot 2021-08-03 at 7 45 37 PM


robin-macpherson:
Woohoo can't wait to take a closer look today 🎉 Regarding the TODO, let's not do that until the work on #617 is finished which should be in the next couple of days

Lanny:
> If this is not desired, then we could create a new user fragment that contains the interests, specifically for the pages where it is needed

I think this would be my preferred approach, just because it reduces the amount of bureaucracy most other pages have to go through. 

Also I’d suggest making the mutations return a user type instead of a topic type (and move them to the user file accordingly), that way there wouldn’t be a need to call ‘refetch()’ after the mutation, since the mutation itself will return an updated user record and Apollo should be able to reconcile the cache on its own, saving a roundtrip to the server. 

That optimization aside, the approach looks good 👍 , good stuff!

robin-macpherson (commented):
Amazing work @meredithcat ! I have the same comments as @Lanny - typically we favor the approach of creating a fragment in this scenario which you'll see a lot of around the codebase as you explore it more. Happy to approve and get this merged in once those optimizations are made, which is really exciting 🥳

meredithcat:
Done - looks like I actually didn't need an extra fragment here - now the `UserInterestFragment` is only being used in the profile & settings pages.

meredithteo:
@robin-macpherson @Lanny Do you all have bandwidth to finish up reviewing this change? If so, I think it is close to being ready for merge. 🚀

robin-macpherson:
@meredithcat I'm drowning in stuff at the moment but we'll definitely find the time to get your two PRs through this coming week. Thanks for taking care of the requested changes here.

packages/web/resolvers/topic.ts
```diff
@@ -119,4 +128,68 @@ const TopicQueries = extendType({
...
+        const interestFilter = {
+          where: {
+            userId_topicId: {
+              topicId: args.topicId,
+              userId,
+            },
+          },
+        }
+
+        const userInterest = await ctx.db.userInterest.findUnique(interestFilter)
```
Lanny, line 183: Is checking for the existence of the user/topic relation. necessary? We fail if it doesn't exist, why not just attempt the delete and fail if the delete doesn't go through? That way we could save a round trip to the DB

Lanny (commented):
Looks good, added one comment where I think we can save a DB query but otherwise seems good to me.

packages/web/resolvers/topic.ts
meredithteo: hmm, this is also done here: https://github.com/Journaly/journaly/blob/master/packages/web/resolvers/language.ts#L139

let me know if you would still advise to remove this check!

Lanny: Yeah, in both cases I think we should just do the delete. `db.delete()` will throw if no record is found for deletion, so we can either just let that exception bubble up or catch and rethrow our own exception, it's the same result either way.

Lanny:
Type checker is complaining (not sure if you can see the circle build logs, but `npm run type-check` or `npm run build` should produce the same message):

```
./pages/dashboard/settings/profile.tsx:43:21
Type error: Type '({ __typename?: "UserInterest" | undefined; } & { __typename?: "UserInterest" | undefined; } & { topic: { __typename?: "Topic" | undefined; } & { __typename?: "Topic" | undefined; } & Pick<...>; })[] | undefined' is not assignable to type 'UserInterest[]'.
  Type 'undefined' is not assignable to type 'UserInterest[]'.

  41 |                   <InterestsForm
  42 |                     topics={data.topics}
> 43 |                     userInterests={data.currentUser?.userInterests}
     |                     ^
  44 |                     refetch={refetch}
  45 |                   />
  46 |                   <SocialForm socialMedia={data.currentUser?.socialMedia} refetch={refetch} />
```

This is kind of an annoying situation (which is not your fault). We're checking that the `currentUser` object returned by `AuthGate` is not undefined, but not checking that `data.currentUser` is not undefined. I don't know that they could ever realistically have different values (maybe if the relevant queries load separately) but TSC can't prove it if so. I think the immediate solution is to check `data.currentUser` along side checking `currentUser` here: https://github.com/Journaly/journaly/pull/633/files#diff-f8b3bc26037afc850349067e596a1f2f4d21e340f5c6f141a5a6d499bbe0b431R29

robin-macpherson:
@Lanny I went ahead and addressed your requested changes re: removing those relation checks and also fixed the build. While testing I also found an existing bug where the selected ID state was not being cleared after adding a language or interest, which resulted in the submit button remaining active and the user could click it again, causing a Unique Constraint error thrown on the server (video below). Each fix in separate commits from today, if you can take a look and let me know if all looks good to you that would be awesome 🙏🏼

https://user-images.githubusercontent.com/34203886/135686673-fca641d9-d318-4cb3-aab0-c31475b10a5d.mov

Lanny (approved):
👍 good stuff

Issue #425, opened 2021-01-06
Better Commenting Experience/Formatting


robin-macpherson:
#### Perceived Problem

- Currently, our comments are incredibly basic and a number of users have requested better options, in particular for our comments to respect line breaks, bold, underline, and italics.

#### Ideas / Proposed Solution(s)

- Respect line breaks inside of comments
- Limited formatting with markdown
  - support bold, italic, underline
  - maybe a tooltip or instructions?

#### Notes

- We may want to implement the same Rich Text editing experience as with our Slate editor in the future, but in the short-term markdown seems like a far simpler solution
- When looking into a library for markdown support, it would be great to find one that support Ruby Text

robin-macpherson:
Just saw this inside of Jira at work while creating a new ticket, as an example of one approach for inserting some simple instructions as placeholder text:

![image](https://user-images.githubusercontent.com/34203886/103803809-f2976a00-5005-11eb-874e-1093547f65e6.png)

davidsonsns:
@robin-macpherson as we talked before, can you assign it to me?

If so, I'd like to know if you guys have some markdown lib in mind? Saw that you mentioned using some lib to implement this feature. But, I do believe that it can be a little weird to follow using different approaches. I was thinking to implement slate here also, but more generic, I mean with fewer features than post body descriptions.

Let me know your thoughts about it.

robin-macpherson:
Hi @davidsonsns , done! 😄

Yes so we were originally thinking that adding a simpler/more generic implementation of our Slate editor here would be the optimal/most desired solution, but it seems like quite a big project so were were thinking that in the short-term some simple markdown support might be a smaller/easier task.

However you've now worked quite a lot on the J-Editor and have gotten quite familiar with it, and so I would say that if you are excited about working on that first-choice solution here, let's do it! 💪🏼🔥

robin-macpherson:
@Lanny @davidsonsns my only concern about implementing a lighter version of J-Editor here with Slate is that we do still have some pretty nasty bugs that are really unfortunate and mostly due to upstream issues that we still need to figure out how to solve for our users... so this solution would mean that all those impacted users (anyone writing in Korean or Chinese, for example) would now potentially experience those bugs in the commenting experience.

However, I think we should be striving for the best solutions whenever possible and this would be inline with our general philosophy of building for the best long-term code quality, functionality, and User Experience which is all great. But I think this needs to be at least brought up before we commit to that solution

davidsonsns:
[...]

PR #635, opened 2021-08-04
Markdown support for post & inline comments

meredithcat:
## Description

**Issue:** #425 

Add Markdown support & basic styles for Markdown elements (bold, italic, headers, bullet points, code, & blockquotes) in post & inline comments.

Let me know if there is any feedback on the styles - I tried to do something reasonable for each, but you may prefer to change some or leave them out entirely (e.g. we could just not style headers if we think it would be too distracting when reading comments, and just render them as regular text).

For blockquotes I used the same color code as the rest of the site, but we could make it lighter or darker instead.

## Subtasks

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Screenshots

Post comment:
Screen Shot 2021-08-03 at 11 14 36 PM

Inline comment:
Screen Shot 2021-08-03 at 11 14 50 PM


robin-macpherson (commented):
Very exciting stuff @meredithcat ! 🐈✨

I'll take a more thorough look at this a little later today but quick/minor feedback is that it looks like you're running a different NPM version to us which is generating a v2 lock file with the 18,000 lines changed. We have the version pinned in our `.nvmrc` at node v14.15.4, npm v6.14.10 at the moment. Would you mind regenerating the lock file with those versions? Again it's super minor and we may upgrade soon if current active LTS release is there, but we like to keep the PRs really clean & specific, and upgrade our versions together just so that each PR doesn't generate a complete new lock file based on each person's local versions with huge changesets. 

Can't wait to take a closer look and for this to go through! 🚀 I'll definitely make plenty of time for this and the other PRs today 💫

meredithcat:
On second thought, I think I would prefer not styling headers, as they are pretty distracting to read and I can't think of many valid uses for them. Or we could just style them all as a little larger (maybe 1.2em) with font-weight bold. If that sounds ok, then I'll make that change in addition to the npm version number.

robin-macpherson:
> On second thought, I think I would prefer not styling headers, as they are pretty distracting to read and I can't think of many valid uses for them. Or we could just style them all as a little larger (maybe 1.2em) with font-weight bold. If that sounds ok, then I'll make that change in addition to the npm version number.

Yeah I agree. I would say maybe let's go with just a tiny bit larger font-size and font-weight for headings, and apply the same styles to all headings. That way if people do want to include them, they can see that we don't ignore/not support headings, but it's also not to disruptive to the comment. Great call! :)

packages/web/public/static/locales/en/comment.json
```diff
@@ -4,7 +4,7 @@
   "save": "Save",
   "cancel": "Cancel",
   "submit": "Submit",
-  "addCommentPlaceholder": "Add a comment...",
+  "addCommentPlaceholder": "Add a Markdown comment...",
```
robin-macpherson, line 7:
```suggestion
  "addCommentPlaceholder": "We now support Markdown in comments!! Try using **bold** or *italic*!",
```

packages/web/components/Dashboard/Post/PostComment.tsx
```diff
@@ -216,6 +217,37 @@ const PostComment: React.FC<PostCommentProps> = ({
...
+        }
+
+        :global(.comment-body code) {
+          background-color: #eee;
+          font-family: monospace;
+          padding: 2px;
+        }
+
+        :global(.comment-body blockquote) {
+          border-left: 2px solid #4391C9;
```
robin-macpherson, line 247:
```suggestion
          border-left: 2px solid ${theme.colors.blueLight};
```

And then `import theme from '@/theme'`

packages/web/components/Dashboard/Post/PostComment.tsx
```diff
@@ -216,6 +217,37 @@ const PostComment: React.FC<PostCommentProps> = ({
           word-wrap: break-word;
         }
 
+        :global(.comment-body h1) {
+          font-size: 1.5em;
+          font-weight: bold;
```
robin-macpherson, line 222:
```suggestion
          font-weight: 600;
```

We generally prefer to specify font-weights, same thing on line 230

packages/web/public/static/locales/en/comment.json
meredithcat: Hmm, this message looks a bit bulky especially with inline comments. I think `Add a Markdown comment...` looks more streamlined but is also not as clear. What do you think?

Screen Shot 2021-08-07 at 10 23 54 PM

meredithcat:
I re-generated `package-lock.json`, and now it is only 4,000 lines changed. Let me know if that is what you had in mind, or if not, if there is another way to re-generate the file using the correct versions.

robin-macpherson (commented):
Thanks for making the changes!

Regarding the placeholder being a little bulky, I think it's fine. In a case like this I'll generally always prefer clarity over looking a bit more streamlined, especially since this is a temporary thing.

Hmm I do see that the build is failing but at first glance the failure looks a bit odd, I'll take a look at that and the package-lock stuff once settling into some Journaly work today

robin-macpherson:
Just to update here re: the build failure situation, while running some builds locally I found some interesting things that I'm hoping to look into today. Will update

robin-macpherson:
Hi @meredithcat ! Ok so after [some very spirited detective work](https://github.com/Journaly/journaly/pull/638), everything has been solved. Could you please:
- merge `master` into this branch
- make sure you're on the pinned Node & NVM versions in our `.nvmrc` file - if you're using NVM (highly recommend) you can simply run `nvm use` from within the project and it will automatically switch you to the correct versions
- regenerate the `package-lock.json` file
And we should be all ready to finally merge this one in! ✨🔥✨🔥✨🔥🚀

Lanny:
Good stuff, excited to get markdown support 👍 

Just wanted to check on images though. CommonMark, as implemented by remark, as used by react-markdown supports inline images (by naming a URL, no upload machinery). I'm not sure we want that inline comments, it seems like a layout nightmare. It looks like react-markdown supports blacklisting elements so we might want to do that for images.

@robin-macpherson do you have strong feelings one way or the other on inline images in comments?

robin-macpherson:
> Just wanted to check on images though. CommonMark, as implemented by remark, as used by react-markdown supports inline images (by naming a URL, no upload machinery). I'm not sure we want that inline comments, it seems like a layout nightmare. It looks like react-markdown supports blacklisting elements so we might want to do that for images.

Ah, great call @Lanny ! Yes I'd say let's not support images within comments, certainly not at the moment until we've had time to give it proper thought 👍🏼

robin-macpherson (approved):
Hi @meredithcat ! Since the remaining work on this one was super minor and it'd be absolutely amazing to get this through I just took care of regenerating the package-lock and disallowing image elements. Thanks so much again for your work on this! 🤟🏼🎸🔥✨

PR #638, opened 2021-08-10
Upgrade Node, Next, React, TypeScript, & Fix Webpack Build Failures

### LINKED ISSUE #530: Upgrade React version to 17.x line
We get this little warning on every build

```
warn  - React 17.0.1 or newer will be required to leverage all of the upcoming features in Next.js 11. Read more: https://err.sh/next.js/react-version
```

It's good housekeeping and it would be nice to do this now instead of waiting until we need to upgrade next.

Probably comes along with some other library upgrades that currently specify a peer dependency of react<17

robin-macpherson:
closes #530 

## Description

Through investigating some new build errors that appeared, I experimented with various package upgrades and eventually found that the issue appeared to be with Next.js now defaulting to using webpack 5 - which seems to have caused various issues trying to resolve server-only dependencies in the client and began giving us trouble with `@prisma` internal dependencies along with some others.

Upon further investigation, the `@prisma` related stuff was addressed with a workaround discussed in [this issue](https://github.com/prisma/prisma/issues/6564) and looks like it will be fixed [in this PR](https://github.com/prisma/prisma/pull/7931). I used the same workaround to address a couple of other internal deps outside of Prisma as well... 🧐

I thought it would be worth keeping all the work I did on finally upgrading to React 17 along with TypeScript and the new major release of Next, along with updating our pinned Node & NPM versions to latest LTS.

Also includes fixing a couple of other things that started failing:
- `node-fetch` dependency & types
- `webhook.ts` issue

Need to check on the impact of the React & Next upgrades with our Next i18n stuff, as it pertains to #543 for example.
Also had trouble updating ESLint to v7 due to things exploding around React and `eslint-plugin-react` so it complains about being on an older version. Would be good to get that upgrade sorted too so everything's in lockstep.

**Other References:**
- https://nextjs.org/docs/upgrading


packages/web/lib/apollo.tsx
```diff
@@ -122,13 +115,9 @@ export function withApollo<PageProps extends object, PageInitialProps = PageProp
...
 
           if (ssrContext.redirectTarget && ctx?.res) {
-            (ctx.res as any).redirect(ssrContext.redirectTarget)
+            ;(ctx.res as any).redirect(ssrContext.redirectTarget)
             return pageProps
           }
-
-          // getDataFromTree does not call componentWillUnmount
-          // head side effect therefore need to be cleared manually
-          Head.rewind()
```
robin-macpherson, line 131: > `Head.rewind has been a no-op since Next.js 9.5, in Next.js 11 it was removed. You can safely remove your usage of Head.rewind.
https://nextjs.org/docs/upgrading

.nvmrc
```diff
@@ -1 +1 @@
-14.15.4
+14.17.4
```
robin-macpherson, line 1: Latest LTS

PR #651, opened 2021-08-18
Dependency Updates & Untangling 🧶

### LINKED ISSUE #648: Fix Versioning Issues
### Description

We have some untangling to do:
- Figure out removing currently externalizes dependencies from Webpack
  - [Relevant PR](https://github.com/prisma/prisma/pull/7931) for `_http_common`
- Next.js versioning: why does it break the build?
- Let's try upgrading `react-markdown` & `remark-gfm` to latest
- ESLint upgrade
  - https://nextjs.org/docs/basic-features/eslint#eslint-config

robin-macpherson:
## Description

**Issue:** closes #648 

We've had quite a few important dependency updates recently, as in #638 , where things got a little bit tangled with certain packages would break thing at a minor version bump occurring in `package-lock.json` (Next), some things just breaking at latest version (ESLint - related to Next), and then `react-markdown` came out with a major update during our PR #635 that also broke the build whereas previous version was fine. This was compounded when trying to install `remark-gfm` at its new latest version (`2.0.0`), likely due to a mismatch in the underlying Remark versions of the two packages.

This PR seeks to untangle everything and get things fixed up at all latest versions 🧶🐈

## Subtasks

- [x] Install latest of everything and see what breaks
- [x] Figure out what in the world is going on with i18n/translation bizarreness! (thank you @Lanny 🙇🏻‍♂️)


robin-macpherson:
Current remaining issues:

A: i18n translations not loading - issue with `getInitialProps` exhibiting strange behaviour/partially complete returned JSON files
Screenshot 2021-08-18 at 15 58 49

Lanny:
Ok, many many commits later and like two lines actually changed I think this fixes our issues.

The immediate issue is that during SSR next-i18next's `fs` calls to get translations off the filesystem failed because these files moved. It seems the files moved because of the next version change, although how exactly vercel deploys a next build artifact is a bit of a mystery so there might be an interaction there.

This change ensures translations get copied into the build and adds logic to properly reference it in the vercel environment.

This is the second time we've lost hours to days because of a next-i18next/vercel impedance mismatch. next-i18next in latest and beyond is dropping support for the current `getInitialProps` fetching mechanism which we're stuck to because of Apollo, and next now has build in locale subpathing so the value prop of the library over just using i18next itself is kind of dubious.

I think we should look at kicking next-i18next as a dependency and just using react-i18next and handling translation fetching ourselves. Next has dynamic import logic which I imagine would be a _much_ more consistent way to handle translation loading, it gets us out of the somewhat hacky `namespacesRequired` approach, would let us clean up our accreted next-i18next hacks, and might fix the longstanding annoyance of translations not getting reloaded during dev.

robin-macpherson:
Amazing work tracking the issue down @Lanny ! I agree with your thoughts on moving forward without next-i18next pretty soon. Seems like a far more stable approach and will be good to further reduce our dependencies and points of failure here. Tested and everything appears to be good to go from what I can see as well! 🚀

PR #639, opened 2021-08-12
Create "Advanced Filters" UI

robin-macpherson:
## Description

**Issue:**
- closes #617 
- closes #452 

Creates an *Advanced Filters* UI for filtering post feeds, and expands these filters to also be implemented on the *My Posts* page as well as the *My Feed* page.

- Refactors all filters into a new `<Filters />` component
- Within `<Filters />`, adds the ability to toggle *Advanced Filters* saving significant screen real estate and paving the way for additional filters to be added

## Subtasks

- [x] I have added this PR to the Journaly Kanban project ✅
- [x] Refactor all search filters to `<Filters />` component
  - [x] Decide correct approach here
- [x] Figure out visual design for advanced filters
- [x] Create Advanced Filters section
- [x] Figure out properly merging `feed` and `posts` queries along with all the required changes throughout the My Posts and Profile pages
- [x] Implement pagination on My Posts page
- [x] Implement pagination on Profile page
- [x] Polish up all the styling so that everything looks right! 💅🏼

### Deployment Checklist

- [ ] 🚨 Publish new j-db-client version and update in both `web` and `j-mail`
- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 DEPLOY `j-mail` to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod
- [ ] 🚨 DEPLOY `j-mail` TO PROD

## Migrations

No Migs

## Screenshots

**A:** Advanced Filters Collapsed
Screenshot 2021-09-09 at 17 53 05

**B:** Advanced Filters Expanded
Screenshot 2021-09-09 at 17 53 16

PR #657, opened 2021-09-11
Fix profile page bug fetching currentUser's posts

robin-macpherson:
## Description

A bug was introduced with #639 where a user is seeing their own posts on another user's profile. This PR fixes that.

## Subtasks

- [x] Fix bug and test locally
- [ ] Test on stage

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

No migs

### DIFF
--- a/packages/web/graphql/language/languages.graphql
+++ b/packages/web/graphql/language/languages.graphql
@@ -1,4 +1,4 @@
-query languages($hasPosts: Boolean, $authoredOnly: Boolean) {
+query languages($hasPosts: Boolean, $authoredOnly: Boolean, $topics: [Int!]) {
   languages(hasPosts: $hasPosts, authoredOnly: $authoredOnly) {
     ...LanguageWithPostCountFragment
   }
--- a/packages/web/graphql/post/posts.graphql
+++ b/packages/web/graphql/post/posts.graphql
@@ -9,6 +9,7 @@ query posts(
   $hasInteracted: Boolean
   $authoredOnly: Boolean!
   $status: PostStatus!
+  $authorId: Int
 ) {
   posts(
     first: $first
@@ -21,6 +22,7 @@ query posts(
     hasInteracted: $hasInteracted
     authoredOnly: $authoredOnly
     status: $status
+    authorId: $authorId
   ) {
     posts {
       ...PostCardFragment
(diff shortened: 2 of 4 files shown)


packages/web/graphql/language/languages.graphql
```diff
@@ -1,4 +1,4 @@
-query languages($hasPosts: Boolean, $authoredOnly: Boolean) {
+query languages($hasPosts: Boolean, $authoredOnly: Boolean, $topics: [Int!]) {
```
robin-macpherson, line 1: Restoring this line was necessary to generate the Apollo hooks at this stage. I remember we removed it when seeing that it didn't appear to be doing anything in the languages query, but it appears to be used in the ObjectType in a way that I don't fully understand, but I think that's why it was here originally.

packages/web/graphql/post/posts.graphql
```diff
@@ -9,6 +9,7 @@ query posts(
   $hasInteracted: Boolean
   $authoredOnly: Boolean!
```
Lanny, line 10: Since we're restoring `authorId` we should get rid of `authoredOnly` shouldn't we?

packages/web/graphql/language/languages.graphql
Lanny: I guess the `postCount` property on languages uses it? But I don't know what that's for. I think we should try to get rid of this if it's not doing anything useful

Lanny (commented):
Left a couple comments, but on looking at this closer there are some issues in the resolver that also need to be addressed.

First one is here: https://github.com/Journaly/journaly/blob/master/packages/web/resolvers/post.ts#L380

this is a strange interaction of argument. I think we should just get rid of `authoredOnly`, I don't think we ever actually need it since we can just pass the current user ID as `authorId`.

The second is around status filtering: https://github.com/Journaly/journaly/blob/master/packages/web/resolvers/post.ts#L388

This allows any logged in user to see other users drafts (passing no `status` argument, or `status` as "DRAFT") which is obviously bad. It also means non-logged in users will see a mix of draft and non-draft posts, regardless of what args they pass. The solution is we should only allow the status filter to be something other than "PUBLISHED" when both a. there is a current user, and b. the `authorId` arg is equal to the current user ID

PR #658, opened 2021-09-12
Rewrite posts query

Lanny:
## Description

- Fixes perf issues with "needs feedback" filter on posts query
- Fixes issue with drafts being revealed when they shouldn't if you post tamper

## Migrations

No migs


Lanny:
@robin-macpherson let's test this in preview first, then let's deploy when we have time for prod smoke testing right after it goes out

robin-macpherson (approved):
Amazing, thanks so much for the perf updates and other wee fixes!

PR #700, opened 2021-11-30
⌨️ J-Editor / FEATURE: Typewriter Sounds! 🎵

### LINKED ISSUE #494: Mobile editor buttons overrun their container
Screen Shot 2021-02-06 at 5 20 16 PM

Seems to happen at most portrait mobile resolutions

### LINKED ISSUE #699: ⌨️ J-EDITOR: Optional Typewriter Sounds
### Description

I've always thought it would be super cool to have a little option you could toggle on and off to include sweet little typewriter sounds with each keystroke as you type! I was just reminded of this when re-starting an online touch typing course where they have this feature and it is legit so satisfying 😌🧘🏻‍♂️

### Screen Recording

https://user-images.githubusercontent.com/34203886/144016839-21f68f0d-2b53-4dd6-a6fa-b5595100ceb6.mov

robin-macpherson:
## Description

**Issue:**
- closes #699 
- fixes #494 

Welcome to a delightful, fun feature that simply adds a sweet and delightful edge to the Journaly writing experience: typewriter sounds! Because why not!? Basically, if a user turns this feature on then they'll simply hear the delightful clickity-clacks of a typewriter with each key press ⌨️ ❤️

Also takes care of some important work that will become useful in other upcoming PRs:
- Builds a `<SwitchToggle />` Component which will be very handy for quickly toggling on and off features
- Fixes our long-standing bug where the mobile editor toolbar overruns its container, implementing a more flexible layout

## Subtasks

- [x] Implement sound effect on typing
- [x] Figure out issue with sounds not playing in quick succession
- [x] Figure out issue with sounds not playing in very quick succession 😅
- [x] Figure out issue with `playSounds` state not updating correctly in some places
- [x] Build SwitchToggle Component
- [x] Implement toggling in the editor
- [x] Fix responsive layout issues
- [x] Decision: Premium or not?
- [x] Translations
- [x] ~Handle refocusing & holding position on formatting~ (Moved to separate ticket)
- [x] Save toggle state (should just be able to drop in `useAutosavedState`)
- [x] Fix slight delay between when sticky kicks in and fixed class gets added
- [x] Rename toolbarRef
- [x] Observer cleanup (when we unmount let's clean up the observer)
- [x] try clean up dashboardConstants
- [x] Some kind of handling for browsers that don't have observer support (wrap in try-catch)
- [x] Figure out the gotcha with typing on mobile 😩

### Deployment Checklist

- [x] 🚨 Deploy code to stage
- [ ] 🚨 Deploy code to prod

## Migrations

No miggies.

### 1. Toolbar with Typewriter Sounds OFF

Screenshot 2021-11-30 at 11 59 06

### 2. Toolbar with Typewriter Sounds ON

Screenshot 2021-11-30 at 11 59 23


packages/web/public/static/locales/de/j-editor.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typewriterSounds": "Typewriter Sounds"
+}
```
robin-macpherson, line 3: @LindasLinguas a tiny translation here 🤓 ⌨️🎵

packages/web/public/static/locales/de/j-editor.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typewriterSounds": "Typewriter Sounds"
```
LindasLinguas, line 2:
```suggestion
  "typewriterSounds": "Schreibmaschinengeräusche"
```

packages/web/public/static/locales/en/j-editor.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typewriterSounds": "Typewriter Sounds"
```
robin-macpherson, line 2: @harisont just thought I'd point this extra translation string out to you, as it will be missing from the files on your branch ✌🏼🤓

packages/web/hooks/useGetWindowSize.ts
```diff
@@ -0,0 +1,45 @@
...
+type WindowSize = {
+  width: number | null
+  height: number | null
+}
+
+const useWindowSize = () => {
+  // Make sure we're client-side / not server-side
+  const isClient = typeof window === 'object'
+
+  const [windowSize, setWindowSize] = useState<WindowSize>({
```
Lanny, line 24: I totally see the logic here, but this is one of those reference equality gotchas. Because you're passing an object literal to `useState` its value will be updated every time the hook is called (because `{} !== {}`, `useState` will see a new value each time it's called) so this will produce unnecessary re-rendering.

The most straight forward  approach would probably be to track window width and height as two different pieces of state, then at the bottom of the hook do `return useMemo(() => { width: windowWidth, height: windowHeight }, [windowWidth, windowHeight])` (where `windowWidth` and `windowHeight` are the two state values).

packages/web/hooks/userIntersectionObserver.ts
```diff
@@ -0,0 +1,56 @@
...
+  rootMargin?: string
+  threshold?: number
+}
+
+const useIntersectionObserver = ({
+  root = null,
+  rootMargin = '0px',
+  threshold = 0,
+}: IntersectionObserverOptions) => {
+  const [observedElementRef, setObservedElementRef] = useState(null)
```
Lanny, line 28: Why use state rather than an actual ref here? Maybe in case the observed element changes? I guess that could happen if the element is conditionally rendered.

You might need `useState<HTMLElement>` (or `HTMLElement | null` or something) to make TSC happy though

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
```diff
@@ -29,176 +29,197 @@ import ToggleMarkButton from './ToggleMarkButton'
...
-  const toolbarRef = useRef<HTMLDivElement>(null)
-  const [isFixed, setIsFixed] = useState(false)
-  // Save the initial distance that the toolbar is from the top of the page
-  const originalElementOffsetTop = toolbarRef?.current?.offsetTop as number
-  // Save height of toolbar so that when it's fixed and removed from document flow,
-  // the parent takes the same height and surrounding content won't collapse
-  const fixedDistanceFromTop = 20
-  const toolbarHeight = toolbarRef.current?.getBoundingClientRect().height || 56
-  const toolbarClasses = classNames('editor-toolbar-container', { 'is-fixed': isFixed })
+  const stickyToolbarOffset = 20
```
Lanny, line 52: Having this and "toolbarStickyOffset" is a bit confusing. Maybe just get rid of this variable and make `toolbarStickyOffset = <condition> ? -92 : -20` below?

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
```diff
@@ -29,176 +29,197 @@ import ToggleMarkButton from './ToggleMarkButton'
...
-    window.addEventListener('scroll', handleScroll)
-
-    return () => {
-      window.removeEventListener('scroll', handleScroll)
-    }
-  }, [isFixed, originalElementOffsetTop])
+
+  const windowSize = useGetWindowSize()
+
+  let toolbarStickyOffset =
```
Lanny, line 62: this can be a const, I don't think it ever gets re-assigned

packages/web/hooks/userIntersectionObserver.ts
```diff
@@ -0,0 +1,56 @@
...
+
+      return () => {
+        if (observedElementRef) {
+          observer.disconnect()
+        }
+      }
+    } catch (e) {}
+  }, [observedElementRef, root, rootMargin, threshold])
+
+  return [setObservedElementRef, isIntersecting]
```
Lanny, line 53: I _think_ you can do `return [setObservedElementRef, isIntersecting] as const` here to satisfy the compiler.

If that doesn't work, try giving this function a return type of `[boolean, LegacyRef<HTMLElement>]`.

The issue here is that TSC is giving this array a type of `(boolean | <useState's return value>)[]` meaning it's saying any item in this array could have either of those  types. We need to assure it that this is a tuple, meaning the first element is a boolean, the second element is a setState callback (which happens to overlap with `LegacyRef`'s signature) and there's no way for them to switch places.

robin-macpherson: Amazing, this worked! Thanks 🚀

Lanny (approved):

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
```diff
@@ -29,176 +29,209 @@ import ToggleMarkButton from './ToggleMarkButton'
...
-        setIsFixed(false)
-        return
-      }
-    }
-  }
+
+  const windowSize = useGetWindowSize()
+
+  // This is to accounnt for the nav which appears at the top on mobile and has a height of 72px
+  const toolbarStickyOffset =
```
Lanny, line 64: since we un-fixed the nav, is this still necessary? Can we make `rootMargin` a constant `-20px 0 0 0`?

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
```diff
@@ -29,176 +29,209 @@ import ToggleMarkButton from './ToggleMarkButton'
...
+
+  const [toolbarObserverRef, toolbarShouldFloat] = useIntersectionObserver({
+    rootMargin: `${toolbarStickyOffset}px 0px 0px 0px`,
+  })
 
   useEffect(() => {
-    window.addEventListener('scroll', handleScroll)
+    const onVisualViewportChange = () => {
+      setViewportsDiff(visualViewport.offsetTop)
+    }
```
Lanny, line 74: I kinda think we should debounce this for performance purposes, maybe with an interval of half or a quarter of a second, for performance purposes. But I don't know how much of an impact that's going to have on the feeling while scrolling. There's already some inevitable latency in there so maybe not too bad?

If you have a chance maybe test it with and without the debounce and see if it makes a big difference. But not a blocker.

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
```diff
@@ -29,176 +29,209 @@ import ToggleMarkButton from './ToggleMarkButton'
...
+                onToggle={onToggleShouldPlayTypewriterSounds}
+              />
+            </div>
+          </div>
+        </div>
+        <style jsx>{`
+          .editor-toolbar-container {
+            position: sticky;
+            // Update this value when visualViewport (diff between viewport & visualViewport)
+            // listen to changes to size of visualViewport
```
Lanny, line 182: Think this was two different comments that never quite got finished. I'd suggest this:

```suggestion
            // Push toolbar down from the top of the logical viewport to the degree that the logical viewport is outside the visual viewport
```

but IDK if that's much clearer to anyone else

Lanny (approved):
This is looking in quite good shape. I thought with all the fiddling with keyboard hacks we'd have some more spaghetti in here but it's pretty lean and clean all things considered.

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
robin-macpherson: Good catch! Removed this and related code ✅

PR #706, opened 2021-12-07
POSSIBLE FIX TO OUR LAST MAJOR J-EDITOR ISSUES

### LINKED ISSUE #479: BUG: Impossible to write or edit Korean text on mobile

robin-macpherson:
## Description

Brings in most recent Slate updates, which contain three very exciting pieces of work, but one in particular: [
Fix CJK IME(like Chinese, Japanese) double input ](https://github.com/ianstormtaylor/slate/pull/4702) which addresses the doubling of composed character input and may finally address the last of our major Slate issues, where some users on certain devices have continued to experience the issue with characters doubling in languages like Japanese, Korean, and Chinese 😱🔥

**Issue:**
- fixes #479 (TBD)

Also improves performance for Android users 📉

## Subtasks

- [x] Test for regressions
- [ ] Testing from users who have still been experiencing the issue until now

### Deployment Checklist

- [ ] 🚨 ~Publish new j-db-client version and update in both `web` and `j-mail`~
- [ ] 🚨 ~Deploy migs to stage~
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 ~DEPLOY `j-mail` to stage~
- [ ] 🚨 ~Deploy migs to prod~
- [ ] 🚨 Deploy code to prod
- [ ] 🚨 ~DEPLOY `j-mail` TO PROD~

## Migrations

No migs

## Screenshots
N/A


robin-macpherson:
@rickyweb !!! VERY exciting possible updates! Are you available to re-test the issues with character duplication in Korean here: https://journaly-git-hugely-exciting-slate-update-journaly.vercel.app

I don't want to get tooooooo excited but I *think* the problem may have FINALLY been solved 🤞🏼😬🤞🏼 This would be hugeeeeeeeeeeeeeeeee if that is the case! 

cc: @LindasLinguas could you perhaps also test on your side just to make sure there are no regressions?

Fingers crossed!!

LindasLinguas:
Hi Robin,
Sadly no improvements on my side. The cursor still jumps to the corrected letter on Android devices.

robin-macpherson:
Hi @LindasLinguas , thanks! Yes there wouldn't be any improvements to the Android/Swiftkey spell check issue here, it should hopefully be specifically the problem with characters that involve character composition (like Chinese characters or Korean letters) still doubling in some languages on some devices that is fixed 🤞🏼 I was just asking for additional testing to make sure that nothing else had gotten worse or no new bugs introduced with these changes.

robin-macpherson:
I'm going to go ahead and merge this to production as it seems there are no regressions or other issues and there is very good reason to believe that these updates appear to directly address and improve major issues that some users might be experiencing each day. I'll try to find some other users to verify because this would honestly be **HUGE** huge huge huge if it does finally address the remaining major issues for all those people 🤞🏼

PR #717, opened 2021-12-11
💅🏼 Prettier Profile URLs & Remove `/dasboard/` URL Pattern

### LINKED ISSUE #629: 🔑 Allow access to a user's profile by handle
#### Perceived Problem

Currently, the URL to a user's profile is linked to their user id and not their handle, giving a URL like `journaly.com/dashboard/profile/8654`. This is not an ideal experience, as users don't typically share their user id with others if they want to share their work outside of Journaly. It would be more convenient for them to share their username.

- Add another URL, maybe something like `journaly.com/user/<handle>`, to access a user's profile by their handle.
- When changing their handle in the settings page, give users a warning that this will invalidate their old profile URL and that they should update it outside of Journaly.
- Change the original URL page so that it looks up a user's handle and then redirects to the new URL, to maintain backwards compatibility.

### LINKED ISSUE #748: 🔑 Prettier Profile URLs - Optimizations
#### Perceived Problem

Our current approach from #717 requires the client load the full page before doing a redirect.

- @Lanny had the good thought that we can do some magic in `getInitialProps` to redirect via status code which would make this a bit cleaner

robin-macpherson:
**Issue:**
- closes #629 
- closes #748 

### Part 1

- Improves the User Experience of visiting and linking to profiles by creating a new API route, `/user/[handle]` to locate the user and load their profile page instead of `/profile/[id]` as it was before.
- Maintains the original profile route which now simply looks up the user and redirects them to the new route for backwards compatibility and not breaking links that have been posted until now.
- Implements a warning when users update their handles that they'll need to update any links to their profile
- We have chosen to NOT store previous handles and maintain old links for the user

### Part 2

- Updates and shortens all Dashboard URLs by completely removing `/dashboard` from them.
- So `journaly.com/dashboard/my-posts` has just become `journaly.com/my-posts`
- All old URLs have been maintained and currently simply redirect users to the new pages for backwards compatibility
- **TODO:** There may be some optimizations we can do to improve the mechanism for redirecting to the new pages

## Approach For Prettier Profile URLs

After discussion, the decided upon approach was to simply expand the `userById` query to also accept a handle and make both `id` and `handle` optional, only allowing one to be passed at a time and then performing the query based on that. I also chose to rename this query to `userByIdentifier` since this is still accurate (as we're passing one of two unique identifiers) and less confusing.

## Subtasks

- [x] Build out initial POC
- [x] Get feedback on the approach & refine
- [x] Handle warning message when a user updates their handle
- [x] Translations
- [x] Refactor all code related to new URL pattern with backwards compatibility
- [x] Test initial work
- [x] Improve About page styling
- [x] FIRST: Look into redirect rules in `next.config.js`
- [x] Verify that we can delete the old page components
- [ ] ~Create Redirect Component~
- [ ] ~Handle redirects in `getInitialProps` (possibly for pages except user)~
- [x] Be sure to check paths that require query params or path matching
- [x] Fix bug with profile page redirects:
  - `Unhandled Runtime Error: Rendered more hooks than during the previous render.`

### Deployment Checklist

- [x] 🚨 Deploy code to stage
- [x] 🚨 Deploy code to prod

## Migrations

No los hay ✌🏼

## Screenshots

Tampoco ✌🏼


packages/web/public/static/locales/de/settings.json
```diff
@@ -31,7 +31,8 @@
       "cityLabel": "Stadt (optional)",
       "cityPlaceholder": "In welcher Stadt lebst du?",
       "countryLabel": "Land (optional)",
-      "countryPlaceholder": "In welchem Land lebst du?"
+      "countryPlaceholder": "In welchem Land lebst du?",
+      "updateHandleWarning": "Warning: updating your handle will break any existing links you have shared to your profile. If you continue, we recommend you update those links to have your new handle"
```
robin-macpherson, line 35: @LindasLinguas there's a new string here to be translated whenever you have time 😄🙏🏼

LindasLinguas:
```suggestion
      "updateHandleWarning": "Warnung: Das Ändern des Benutzernamens führt dazu, dass die alten Links zu deinem Profil, die du bereits geteilt hast, nicht mehr funktionieren werden. Wenn du deinen Benutzernamen trotzdem ändern möchtest, empfehlen wir diese Links zu aktualisieren."
```

packages/web/components/Dashboard/Nav/Nav.tsx
```diff
@@ -130,7 +130,10 @@ const Nav: React.FC<Props> = ({ expanded, collapse, disableLargeNav }) => {
           {currentUser && (
             <>
               <div className="nav-top">
-                <Link href={`/dashboard/profile/[id]`} as={`/dashboard/profile/${currentUser.id}`}>
+                <Link
+                  href={`/dashboard/user/[handle]`}
+                  as={`/dashboard/user/${currentUser.handle}`}
```
Lanny, line 135: I see we had the ‘as’ here before but we should be able to get rid of it, just use the href field. 

Also I’d personally like to get “dashboard” out of that URL but that’s just me.

packages/web/components/Dashboard/Post/PostAuthorCard.tsx
```diff
@@ -58,7 +58,7 @@ const PostAuthorCard: React.FC<PostAuthorCardProps> = ({ author }) => {
   return (
     <div className="container">
       <div className="author-info-container">
-        <Link href={`/dashboard/profile/[id]`} as={`/dashboard/profile/${author.id}`}>
+        <Link href={`/dashboard/user/[handle]`} as={`/dashboard/user/${author.handle}`}>
```
Lanny, line 61: Same comment as above here

packages/web/components/Dashboard/Post/PostComment.tsx
```diff
@@ -79,7 +79,7 @@ const PostComment: React.FC<PostCommentProps> = ({
     <div className="comment" id={`pc-${comment.id}`}>
       <div className="author-body-container">
         <div className="author-block">
-          <Link href={`/dashboard/profile/[id]`} as={`/dashboard/profile/${comment.author.id}`}>
+          <Link href={`/dashboard/user/[handle]`} as={`/dashboard/user/${comment.author.handle}`}>
```
Lanny, line 82: “as”

packages/web/components/PostHeader/PostHeader.tsx
```diff
@@ -65,7 +65,7 @@ const PostHeader: React.FC<PostHeaderProps> = ({
           <p> &mdash; </p>
           <p>
             {t('postBy')}{' '}
-            <Link href={`/dashboard/profile/[id]`} as={`/dashboard/profile/${authorId}`}>
+            <Link href={`/dashboard/user/[handle]`} as={`/dashboard/user/${authorHandle}`}>
```
Lanny, line 68: “as”

packages/web/components/UserList/UserList.tsx
```diff
@@ -13,7 +13,7 @@ const UserList: React.FC<UserListProps> = ({ users, colorScheme = 'light-mode' }
   return (
     <div className="list-container">
       {users.map((user) => (
-        <Link href={`/dashboard/profile/[id]`} as={`/dashboard/profile/${user.id}`} key={user.id}>
+        <Link href={`/dashboard/user/[handle]`} as={`/dashboard/user/${user.handle}`} key={user.id}>
```
Lanny, line 16: “as”

packages/web/pages/dashboard/profile/[id]/index.tsx
```diff
@@ -1,41 +1,58 @@
...
-      <DashboardLayout pad="never">
-        {userById && posts && (
-          <Profile isLoggedInUser={currentUser?.id === userId} user={userById} posts={posts} />
-        )}
-      </DashboardLayout>
-    </LoadingWrapper>
-  )
+  const { data, loading, error } = useUserByIdentifierQuery({ variables: { id: userId } })
+
+  if (data?.userByIdentifier && typeof window !== 'undefined') {
```
Comment (Lanny) [inline packages/web/pages/dashboard/profile/[id]/index.tsx:15]: This approach requires the client load the full page before doing a redirect. I think we can do some magic in getInitialProps to redirect via status code which would make this a bit clearer IMO. We can merge like this, but that is a follow up I’d like to do

robin-macpherson:
@Lanny thanks for catching/reminding me about those "as" props. I just went through and removed those as well as others that we needed to get rid of.

Regarding removing `dashboard` from the URL, the reason I left it for now is that it feels like a change we'd want to make wholesale across the app at once since users will need to learn this new URL pattern, and it will also break all external links unless we build a mechanism to support the old pattern that redirects to the new one. It felt like something that would need some brainstorming and discussion, WDYT about revisiting that when we're able to have a proper chat about it after the holidays?

Regarding your follow-up idea, I think that's an excellent thought! I can make the follow-up ticket for it ✅

robin-macpherson:
@Lanny just an update, I'm thinking at this point it may just be better wait until we're next able to meet and to do everything at once, so updating our URL patterns across the board (removing `/dashboard` and handling any repercussions of that ex: the current dashboard redirect) along with the optimizations you thought about that I put in #748 . Since the timing won't make much difference in the grand scheme of things, and then our users only need to learn the new patterns once all together. WDTY?

Lanny (approved):

Issue #710, opened 2021-12-09
Layout Improvements For Multiple Actions (Ex: Post Publishing)


robin-macpherson:
#### Perceived Problem

- We've built support for a variety of different actions at this point, especially when it comes to saving a post and what visibility it should have: Draft? Share Privately? Publish Publicly?
- This is FANTASTIC! But we've definitely run into U.I. problems and need a way to make this easier for ourselves in terms of layout, but also making the interface feel more clean and minimalistic for our users 🧘🏻‍♂️🍃

#### Ideas / Proposed Solution(s)

- I think the correct approach will definitely involve some kind of menu that displays the different possible actions
- I'm including a screen recording that I took for a different issue (image editing) but at the end shows a really nice and clean implementation from LinkedIn that could be a great reference for our own use case:

https://user-images.githubusercontent.com/34203886/145417975-6815b9cb-b79e-4107-87d4-c653c584b9f6.mov

PR #723, opened 2021-12-12
🌁  FEATURE:Unsplash Integration

dom-waterson:
(WIP) - Putting this up now to show progress. 

So far got a hacked together prototype, where the user can search unsplash and select a picture for their post.

Need to supply access key shown in the .env.example
Screenshot 2021-12-12 at 18 00 59


robin-macpherson:
Awesome, thanks for the Draft PR @dom-waterson ! Exciting to see the progress on this. I know it's a WIP, would you like us to give it an initial review, or would you prefer that we wait until you feel it's further along? If there's any particular part you'd like review on just let us know and we're happy to do that.

Cheers and keep up the great work! 💪🏼

robin-macpherson:
FYI as a heads up and thinking out-loud, I'll probably work on #710 pretty soon, which I think will also be a great option for the Unsplash integration as well. That way we can have just one button that opens a little menu with the option to upload or find an image on Unsplash.

For now I think the way you've done it works just fine, though, so I'll keep you posted and if I finish #710 before this is done, we can use that approach here, otherwise it will be an easy followup ticket.

packages/web/components/Modals/UnsplashModal/UnsplashModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+          text-decoration-line: underline;
+          color: #b3b3b3;
+        }
+      `}</style>
+    </>
+  )
+}
+
+const UnsplashSearch: React.FC<{ onImageSelect: any }> = ({ onImageSelect }) => {
+  const [data, setPhotosResponse] = React.useState<any>()
```
Lanny, line 67: unsplash-js is a typescript project so I think it should come with some type declarations we can use instead of `any` here and the custom photo type above. Even if they don't export the types we need directly (which it kinda looks like they don't, and that's regrettable) we can do something like:

```
type Awaited<T> = T extends PromiseLike<infer U> ? Awaited<U> : T
type SearchResponse = Awaited<ReturnType<ReturnType<typeof createApi>['search']['getPhotos']>>
```

and use `SearchResponse` as our type here

packages/web/components/Modals/UnsplashModal/UnsplashModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+        .li {
+          margin-bottom: 12px;
+        }
+      `}</style>
+    </>
+  )
+}
+
+const UnsplashModal: React.FC<Props> = (props) => {
```
Lanny, line 111: Typically the approach we take (and there may be an exception to t his somewhere that we should probably fix) is that modals render on the screen if they're rendered by their parent (i.e. no `show` prop) and the parent opens/closes the modal by doing the conditional rendering itself, rather than always rendering an passing in a prop indicating that the child should conditionally render.

packages/web/components/Modals/UnsplashModal/UnsplashModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  height: number
+  urls: { full: string; regular: string; raw: string; small: string; thumb: string }
+  color: string | null
+  user: {
+    username: string
+    name: string
+  }
+}
+
+const unsplashApi = createApi({
```
Lanny, line 25: You don't have to do this (although you're welcome to if you'd like), this is mostly a note to self, but I'd like  to dynamically import either this library or or the `UnsplashModal` component itself. The editor page is already one of the heaviest on the site site (which is understandable, the editor involves a lot of code, both from us and third party stuff) but since this is a modal it's a prime candidate to lazy load to cut down on the initial load size of the page.

packages/web/hooks/useImageUpload.ts
```diff
@@ -59,8 +60,22 @@ const useImageUpload = <T extends BaseUploadData>(getUploadData: () => Promise<T
     return result.value
   }
 
-  return [image, uploadingImage, onFileInputChange, () => setImage(null)]
-}
+  const onUnsplashSelect = (smallUrl: string, largeUrl: string) => {
```
Lanny, line 63: `useImageUpload` is shared between avatar, inline, and post image upload flows of which only post image upload involves unsplash integration so this probably isn't the place to add it. 

Also the type of `image` is generic here (because this hook is shared), which is why you need the `@ts-ignore` below.

I think the change you want to make here is to return `setImage` (as the last member of the return tuple so we don't have to refactor every consumer of this hook) and  then either in `usePostImageUplaod` or in the editor itself call `setImage` directly once the user has selected an item from unsplash.

Lanny (commented):
Hey @dom-waterson , thank you for the PR! Unsplash integration is going to be a super nice QoL thing for people who like to always add a post image, and hopefully lowers the barrier to doing so so we get some nice image diversity in the feed. Looking forward to it!

I left a handful of comments. I see this is still a draft so feel free to ignore them for now if you're still iterating on a design. Only thing that might change the direction you're going in a bit is the last one, on `useImageUpload`

packages/web/components/Modals/UnsplashModal/UnsplashModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+  )
+}
+
+const UnsplashModal: React.FC<Props> = (props) => {
+  return props.show ? (
+    <>
+      <Modal
+        title="Search Unsplash"
+        body={<UnsplashSearch onImageSelect={props.onUnsplashSelect} />}
+        onClose={props.onCancel}
```
robin-macpherson, line 117: We generally favour destructuring props

packages/web/components/PostEditor/PostEditor.tsx
```diff
@@ -224,6 +228,16 @@ const PostEditor: React.FC<PostEditorProps> = ({
           topics={postTopics}
         >
           <div className="header-preview-options">
+            <Button
+              type="button"
+              variant={ButtonVariant.Secondary}
+              className="unsplash-button"
+              onClick={(): void => {
```
robin-macpherson, line 235: You should be able to remove these `(): void => ...` type annotations in these inline functions. Generally we let TypeScript figure things out where it can reliably interpret things

robin-macpherson (commented):
@Lanny's review was pretty thorough with some excellent points at this stage so I just went through and left a few comments that are just minor style-guide things, nothing major. Things are looking good overall so far and thanks for including the translations out of the gate ✌🏼

robin-macpherson:
Hey @dom-waterson , just wanting to check in on this. Are you confident you can still make time to work on this feature to completion? If not it's absolutely no problem and I can take over from here, I just wanted to check in because it's been about 3.5 months since you started work on it and this is one I'd really like to get into the product. Just please let me know and happy new year to you!

dom-waterson:
Hey @robin-macpherson sorry about the very late reply! Unfortunately I don't have any time to work on this right now. 😞 Look forward to seeing it in the product. Hopefully soon I can help out elsewhere when I have more time.

PR #728, opened 2021-12-15
Better thread creation experience

### LINKED ISSUE #628: Inline comment box persists even after deleting comment
#### Description

As in the screenshot, we still show the highlight & pop-up comment box even for deleted comments. The desired behavior would be to not show it anymore as it could be confusing for the user.

Screen Shot 2021-08-01 at 8 30 19 AM

#### Repro Steps

Leave an inline comment, then delete it.

### LINKED ISSUE #688: 🔎 Autofocus comment input when opening a thread
#### Perceived Problem

- Currently, if a user creates a highlight to open a thread, once it is created and opens they have to click inside of it

- Let's autofocus into the input once a comment thread opens

### LINKED ISSUE #689: 📉 Improve Performance On Creating/Opening New Thread
#### Perceived Problem

- At the moment, particularly in production but it is also detectable in development, there is quite a delay between clicking the pencil button and the thread actually opening

- Are there ways we can improve performance of the mutation itself?
- Could we implement optimistic UI here?

Lanny:
## Description

**Issue:**
- Fixes #689 
- Fixes #628 
- Fixes #688 

Handful of thread creation tweaks:
- Client now holds thread data until a comment is made, bailing out of a new thread will never touch the DB
- Accordingly opening new thread no longer needs to wait for network IO before the popover  opens
- Thread creation and creating the first comment in a thread is now consolidated into a single call
- Visual indication of thread/comment creation loading state (spinner on submit button)
- `Thread` component now uses refs rather than state to interact with the comment textarea, eliminating re-rendering of the thread component while typing
- Deleting the last comment in a thread will now automatically delete the thread
- Opening a thread will now focus the comment textarea
- "Cancel" button is now always available in a thread, no longer deletes the thread as the thread either isn't created yet or if a thread is empty it will be auto-deleted

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

Includes one migration to allow deleting threads which have subscriptions against them (but only when they're empty). Migration should technically be applied before code, but this isn't critical. No downtime expected.


packages/web/components/Dashboard/Post/Post.tsx
```diff
@@ -457,22 +453,22 @@ const Post = ({ post, currentUser, refetch }: PostProps) => {
...
+      setDisplayCommentButton(false)
+      setPendingThreadData({
+        postId: post.id,
+        startIndex,
+        endIndex,
+        highlightedContent,
       })
+
+      /*
+      */
```
robin-macpherson, line 471: Empty comment

robin-macpherson (approved):
Looks great man! A thought for a potential future improvement is I wonder if we can implement some auto-saving functionality for inputs within threads 🧐 For example if someone clicks out of the popover accidentally and currently loses their progress. We could perhaps store the `highlightedContent` and check if the `highlightedContent` of a new thread contains one that was previously in progress, and if so apply that to the input, clearing it on a comment submission and perhaps even once leaving/re-entering the page so that it's a relatively short-term backup primarily just for those cases of accidentally closing the popover. WDYT?

Anyway, so glad to get these improvements in! 🦔🚀

Issue #604, opened 2021-06-16
🚪 Handle logged out user experience on My Feed page


robin-macpherson:
#### Perceived Problem

- Originally, we did not include this to be protected by `AuthGate` so that we could potentially allow users to see the feed and experience it before signing up, just as we allow them to view a post before signing up
- currently someone trying to access the pretty straightforward URL `https://journaly.com/dashboard/my-feed` results in a 500 due to missing `currentUser` data used in the queries
- Currently, we don't handle this at all and it results in a big Vercel `FUNCTION INVOCATION FAILED` error page, which may come across like a mistake/bug/accident on our part

#### Ideas / Proposed Solution(s)

- I think that we should either build support for this so that an anonymous user can check out the feed page, or decide for now that we are not supporting this, and protect this page with `AuthGate` for a smoother experience

PR #740, opened 2021-12-18
Allow unauth'd users to view profile pages

### LINKED ISSUE #660: Cannot view profiles when logged out
#### Description

Often, I want to share my Journaly profile with someone who doesn't have an account, like my in-laws who are not so tech savvy but want to read my journal entries. It looks like this is currently broken, and shows the error: `An error occurred on the server`

#### Repro Steps

1. Log out.
2. Visit any profile page.

Lanny:
## Description

**Issue:**
- Fixes #660 
- Related to #604 

Fixes issue in resolver where unauth'd user could not use the `posts` query, despite that query being designed for both auth'd and unauth'd cases.

### Deployment Checklist

- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy code to prod

## Migrations

No migs

## Screenshots
Screen Shot 2021-12-18 at 2 27 32 PM

### DIFF
--- a/packages/web/resolvers/post.ts
+++ b/packages/web/resolvers/post.ts
@@ -293,14 +293,16 @@ const PostQueries = extendType({
       resolve: async (_parent, args, ctx) => {
         const { userId } = ctx.request
 
-        const currentUser = await ctx.db.user.findUnique({
-          where: {
-            id: userId,
-          },
-          include: {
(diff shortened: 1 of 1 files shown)


robin-macpherson (approved):
Weeeeehoo! 🚀

Issue #746, opened 2021-12-22
Build First Moderation Tools


robin-macpherson:
## Background

- We currently have two amazing moderators, Linda and Emily, but unfortunately we haven't been able to give them any particular tools yet and so in order to take any action outside of advising a user on the platform they need to go through me.
- This is of course slower, it defeats part of the purpose of having moderators, and it also may make them feel like they don't actually have much additional power, responsibility, or trust as moderators.

## Proposal

- Let's build some basic admin tools that can be used for some basic user and content management
- These should of course be behind a protected route that requires `userRole` of `MODERATOR` or `ADMIN` to access, and then certain actions can be made available to the two roles
- Would like to discuss what the best first actions we can make available should be. Some initial ideas:
  - Flag a post for admin review (spam, hate speech, or other breaches of the WIP community guidelines/terms of use)
  - Flag a comment or post comment for the same reasons

PR #747, opened 2021-12-22
Build JAdmin: Journaly Admin Tools

### LINKED ISSUE #554: Rebuild "Jadmin"
#### Perceived Problem

- Currently, a lot of times we end up needing to manipulate data directly through the database (adding languages, manually updating a user's status, etc.)
- This is potentially error-prone
- Previously, I had built a nice and simple admin interface accessible to administrators within the Journaly app but this was in the old codebase and was not brought over. The entire codebase is of course much different now but, still, high-level design and patterns are already in place

- Rebuild Jadmin!
- Not only is this more convenient, it allows us to make sure that database operations are actually using our code paths which should hopefully be less error-prone, improves testability, but also this could be extended to be used by moderators for a specific set of actions (un-publishing a potentially problematic post/comment/etc for review, for example)

robin-macpherson:
## Description

**Issue:**
- closes #554 
- #746

(Currently a POC)
Builds an initial set of admin tools: JADMIN ✨

## Subtasks

- [x] Build out the initial POC
- [ ] ...
- [ ] ...

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

## Screenshots

image

### DIFF
--- a/packages/web/components/Dashboard/JAdmin/JAdmin.tsx
+++ b/packages/web/components/Dashboard/JAdmin/JAdmin.tsx
@@ -1,0 +1,107 @@
+import React from 'react'
+import theme from '@/theme'
+import { useUsersQuery, UserWithLanguagesFragmentFragment as UserType } from '@/generated/graphql'
+import Button, { ButtonVariant } from '@/components/Button'
+
+type UserInfoProps = {
+  user: UserType
+}
+
+const UserInfo: React.FC<UserInfoProps> = ({ user }) => {
+  return (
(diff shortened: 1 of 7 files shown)


robin-macpherson:
@Lanny just FYI this was intended to be a (very WIP) Draft, but accidentally opened it and it seems you can't go back after that 😑🐈

Issue #574, opened 2021-05-14
BADGES: UI & UX Improvements with Snackbars & Icons!


robin-macpherson:
#### Perceived Problem

- We have the wonderful concept of badges and I think users love this!
- Currently, each individual badge contains the same visual treatment so does not stand out, and currently they don't look like badges visually
- The description of a given badge can only be seen by hovering with the mouse and waiting for the alt text to appear, which is not obvious
- When a badge is achieved, the visibility of this to the user is low. We send an email which is great, but this should also be re-vamped

#### Ideas / Proposed Solution(s)

- Let's update the visual treatment so that each badge gets its own icon that matches the badge criteria and makes each one stand out
- Let's make it more intuitive to see a badge's description, I'm thinking that as a really simple design solution that might be slightly more intuitive and clear is to fade in/out the description on hover
- Let's display a nice little Snackbar when the user earns the badge! See screenshot of an example
- Let's revamp the email design to include our overall aesthetic of the daily digest email but also to include the badge icon

#### Screenshots

<img width="406" alt="Screenshot 2021-05-14 at 07 14 27" src="https://user-images.githubusercontent.com/34203886/118231412-04db5600-b487-11eb-873b-8298ad79f822.png">

robin-macpherson:
Looking forward to this @Lanny ! 🙌🏼

robin-macpherson:
As a little update, now that we have our beautiful Notification Feed, we could create a nice little notification for earning a new badge, too! 😄

PR #752, opened 2021-12-29
🎖 BADGES: UI & UX Improvements

### LINKED ISSUE #525: More badges for achievements
I have seen very, very few badges for achievements - only the `Beta User` and `10 posts`. It would be nice if these badges could be fleshed out a bit more, especially as the platform ages and some users write far more than 10 posts.

I'm aware there are also Alpha + Code Contributor badges, but the percentage of users with these is very low and will only decrease with new sign-ups. The jump from 10 posts to 100 posts is a massive one, and I think it would be reasonable to have some more in-between # of post badges such as 25, 50, and 80.

Other badge ideas include:
- writing streak badges
- number of languages written in
- badges for long posts

I hope we can soon see this feature come to Journaly.

robin-macpherson:
## Description

**Issues:**
- closes #525 
- contributes to #574 

Let's get this badge party started... with ***35 new awesome badges***! 🕺🏻🎖🥇🥈🥉🏅💃🏻

This is the first major part of the work for #574 - completely transforming our User Experience around badges, which I think will be a major part of realizing our Healthful Habit-Forming Design Vision.

This work adds a pretty huge array of badges across a number of different categories: misc, posts written, posts corrected, thanks received, feedback given, and even our first "goal" badge which can't be used yet but shows the direction those will go in.

Includes a migration and also a (TODO) backfill to award users who have already fulfilled the badge criteria for some or even all badges.

TODO: How shall we notify all of those users? Ideally, we could list out each badge they have earned in a single email. 

## Subtasks

- [x] Schema updates & generate mig
- [x] Regenerate client-side GQL `BadgeType`
- [x] Update `Badge.tsx` to support all new badge types
- [x] Write SQL for all new badges
  - [x] Posts Written
  - [x] Comments Given
  - [x] Thanks Received
  - [x] Misc
  - [x] Posts Corrected
  - [x] ~Word count~
- [x] Update badge translation files
- [x] Move badge translations to their own file
- [x] Handle English badge name > enum mapping
- [x] Go over & discuss all new badges
- [x] Discuss translation copy & language
- [x] Discuss badge naming conventions/system
- [x] Look into addressing `db.raw` vs. updating it to `db.queryRaw` now that we have upgraded Prisma (see [this comment](https://github.com/Journaly/journaly/blob/009a94010bd2937f62fdbb7aaa4833e8640756cd/packages/web/resolvers/post.ts#L70))
- [x] Discuss badge accomplishment emails
- [x] Backfilling
- [x] Fix all build errors and stuff

### Deployment Checklist

- [x] 🚨 Deploy migs to stage
- [x] 🚨 Deploy code to stage
- [x] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod
- [x] 🚨 DEPLOY `j-mail` TO PROD

## Migrations

## Screenshots

image


packages/web/resolvers/comment.ts
```diff
@@ -22,9 +22,80 @@ import {
...
   ThreadWithRels,
+  sendNewBadgeEmail,
 } from './utils'
 import { NotFoundError } from './errors'
 
+const assignCommentCountBadges = async (db: PrismaClient, userId: number): Promise<void> => {
+  // Use a raw query here because we'll soon have a number of post count
+  // badges and we could end up with quite a bit of back and fourth
+  // querying, whereas here we can just make one roundtrip.
+  const newBadgeCount = await db.$executeRaw`
```
Lanny, line 33: Now that we're doing this ins two places (and we  may have more in the future) we might consider making this a little more generic. Maybe something like (and I haven't  tested this):

```ts
const joinSqlFragments = (parts: WhateverSqlType[], delimiter: WhateverSqlType) => {
  let joinedFragment = parts[0] || Prisma.empty
  for (let i = 1; i < parts.length; i++) {
    joinedFragment = Prisma.sql`${joinedFragment}\n${parts[i]}`
  }
  return joinedFragment
}

const assignCountBadges = (db: PrismaClient, userId: number, countQuery: WhateverSqlType, tiers: { [key: number]: BadgeType }) => {
  return db.$executeRaw`
    WITH count_query AS (${countQuery})
    INSERT INTO "UserBadge" ("type", "userId") (
      ${joinSqlFragments(
        Object.entries(tiers)
          .map(([threshold, badge]) => Prisma.sql`
            (
              SELECT
                ${badge}::"BadgeType" AS "type",
                ${userId}::integer AS "userId"
              FROM count_query WHERE count_query.count >= ${threshold}
            )
          `)
        )
      }
    )
    ON CONFLICT DO NOTHING
  `
}

const commentTiers = {
  10: BadgeType.TEN_COMMENTS,
  50: BadgeType.FIFTY_COMMENTS,
  ...
}

const commentCountQuery = Prisma.sql`
        SELECT COUNT(*) AS count
        FROM "Comment"
        WHERE
          "authorId" = ${userId}
`

assignCountBadges(db, userId, commentCountQuery, commentTiers)
```

where `WhateverSqlType` is whatever the type of ```Prisma.sql`` ``` is.

Typing all this out I realize it's a bit complicated so we can go over it together at some point if you want, but there's a lot of commonality here so it'd be nice to unify it. Also the `joinSqlFragments` func would be useful in the glob or raw SQL we're doing in the posts resolver too.

robin-macpherson:
@Lanny perhaps we can also add a notification type for earning a new badge 😄

packages/j-db-client/prisma/schema.prisma
```diff
@@ -477,13 +477,53 @@ enum UILanguage {
...
+  FIFTY_POSTS
+  SEVENTYFIVE_POSTS
+  ONEHUNDRED_POSTS
+  ONEHUNDREDFIFTY_POSTS
+  TWOHUNDRED_POSTS
+  TWOHUNDREDFIFTY_POSTS
+  NOVELIST // Wrote a post over X num words
+  MULTILINGUAL // Write posts in 2 languages
+  POLYGLOT // Wrote in X num languages
+  HELPER // Wrote a post in their native language
```
robin-macpherson, line 499: Possible alternative: Blogger?

packages/j-db-client/prisma/schema.prisma
```diff
@@ -477,13 +477,53 @@ enum UILanguage {
...
+  CORRECT_TWOHUNDREDFIFTY_POSTS
+  // Thanks
+  TEN_THANKS
+  FIFTY_THANKS
+  ONEHUNDRED_THANKS
+  TWOHUNDREDFIFTY_THANKS
+  FIVEHUNDRED_THANKS
+  ONETHOUSAND_THANKS
+  ONETHOUSANDTWOHUNDREDFIFTY_THANKS
+  ONETHOUSANDFIVEHUNDRED_THANKS
```
robin-macpherson, line 522: Separate badge for "giving" thanks - someone who's very grateful

packages/j-db-client/prisma/migrations/20211229072205_first_major_badge_run/migration.sql
```diff
@@ -0,0 +1,36 @@
...
+ALTER TYPE "BadgeType" ADD VALUE 'ONEHUNDREDFIFTY_POSTS';
+ALTER TYPE "BadgeType" ADD VALUE 'TWOHUNDRED_POSTS';
+ALTER TYPE "BadgeType" ADD VALUE 'TWOHUNDREDFIFTY_POSTS';
+ALTER TYPE "BadgeType" ADD VALUE 'TEN_COMMENTS';
+ALTER TYPE "BadgeType" ADD VALUE 'FIFTY_COMMENTS';
+ALTER TYPE "BadgeType" ADD VALUE 'ONEHUNDRED_COMMENTS';
+ALTER TYPE "BadgeType" ADD VALUE 'TWOHUNDREDFIFTY_COMMENTS';
+ALTER TYPE "BadgeType" ADD VALUE 'FIVEHUNDRED_COMMENTS';
+ALTER TYPE "BadgeType" ADD VALUE 'ONETHOUSAND_COMMENTS';
+ALTER TYPE "BadgeType" ADD VALUE 'CORRECT_TEN_POSTS';
```
Lanny, line 22: What's the definition of "correcting a post"? Like leaving either at least one comment or inline comment on it?

robin-macpherson: I'm thinking perhaps leaving an inline comment? Since typically an inline comment is more directly related to a correction, whereas a lot of post comments might be general thoughts/discussion or encouragement ("Nice job!). WDYT?

packages/web/resolvers/utils/email.ts
```diff
@@ -154,8 +156,9 @@ const getBadgeName = (badgeType: BadgeType): string => {
     case BadgeType.ODRADEK:
       return 'Odradek'
   }
-
-  return assertUnreachable(badgeType)
+  // TODO: don't bail out here
+  return 'NAME_PENDING'
```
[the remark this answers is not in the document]
Lanny: Or pull them from the translation files

packages/j-db-client/prisma/schema.prisma
robin-macpherson: @Lanny let's add this one if it's not already in there

packages/web/resolvers/thanks.ts
```diff
@@ -85,6 +91,93 @@ const ThanksMutations = extendType({
...
+            2250: BadgeType.TWOTHOUSANDTWOHUNDREDFIFTY_THANKS,
+            2500: BadgeType.TWOTHOUSANDFIVEHUNDRED_THANKS,
+            5000: BadgeType.FIVETHOUSAND_THANKS,
+            10000: BadgeType.TENTHOUSAND_THANKS,
+          }
+        )
+
+        const newThanksGivenBadgeCount = await assignCountBadges(
+          ctx.db,
+          comment.author.id,
```
Lanny, line 129: think this is the bug mentioned on slack, this needs to be `userId` as that's the party that will get the badge assigned to them (even though we're querying against another user's actions).

packages/web/resolvers/thanks.ts
```diff
@@ -85,6 +91,93 @@ const ThanksMutations = extendType({
...
+
+          await Promise.all(
+            newBadges.map((badge) => {
+              return sendNewBadgeEmail({
+                badgeType: badge.type,
+                user: badge.user,
+              })
+            }),
+          )
+        }
```
Lanny, line 179: We can get some better DB parallelism here by doing the two badge assignments in parallel. I'd recommend factoring these out into functions, e.g. :

```tsx
      const assignThanksCountBadges = async (userId: number) => {
        const thanksCountQuery = Prisma.sql`
          SELECT COUNT(*) AS count
          FROM "CommentThanks" AS ct
          JOIN "Comment" AS c
            ON ct."commentId" = c.id
          WHERE c."authorId" = ${userId}
        `

        const newBadgeCount = await assignCountBadges(
          ctx.db,
          userId,
          thanksCountQuery,
          {
            10: BadgeType.TEN_THANKS,
            50: BadgeType.FIFTY_THANKS,
            100: BadgeType.ONEHUNDRED_THANKS,
            250: BadgeType.TWOHUNDREDFIFTY_THANKS,
            500: BadgeType.FIVEHUNDRED_THANKS,
            1000: BadgeType.ONETHOUSAND_THANKS,
            1250: BadgeType.ONETHOUSANDTWOHUNDREDFIFTY_THANKS,
            1500: BadgeType.ONETHOUSANDFIVEHUNDRED_THANKS,
            2250: BadgeType.TWOTHOUSANDTWOHUNDREDFIFTY_THANKS,
            2500: BadgeType.TWOTHOUSANDFIVEHUNDRED_THANKS,
            5000: BadgeType.FIVETHOUSAND_THANKS,
            10000: BadgeType.TENTHOUSAND_THANKS,
          }
        )

        if (newBadgeCount) {
          const newBadges = await ctx.db.userBadge.findMany({
            where: { user: { id: comment.author.id } },
            include: { user: true },
            orderBy: { createdAt: 'desc' },
            take: newBadgeCount,
          })

          await Promise.all(
            newBadges.map((badge) => {
              return sendNewBadgeEmail({
                badgeType: badge.type,
                user: badge.user,
              })
            }),
          )
      }
    }

    const assignThanksGivenCountBadges = async (userId: number) => {
      ....
    }

    ...inside resolver
    await Promise.all([
        assignThanksCountBadges(comment.author.id),
        assignThanksGivenCountBadges(userId),
    ])
```

Note that while both proposed functions here take in an arg they call `userId` it's important we call them with the appropriate user's ID (id of the comment author, not the requesting user, in the thanks received case)

packages/web/components/Badge/Badge.tsx
```diff
@@ -38,26 +181,23 @@ const getBadgeDescription = (badgeType: BadgeType): string =>
...
+  const { t } = useTranslation('badge')
 
   return (
     <>
       <div className="badge" title={t(getBadgeDescription(badge.type))}>
-        <div className="badge-icon" />
-        <span className="title">{t(getBadgeTitle(badge.type))}</span>
+        {/* <div className="badge-icon" />
+        <span className="title">{t(getBadgeTitle(badge.type))}</span> */}
+        <img src={`/images/badges/${badge.type}.svg`} alt="" />
```
Lanny, line 191: Can we use the badge title as the alt text? Does that interact with the `title` attribute on the div strangely? I think providing _some_ kind of alt test here would be nice for a11y purposes.

packages/web/components/Badge/Badge.tsx
```diff
@@ -38,26 +181,23 @@ const getBadgeDescription = (badgeType: BadgeType): string =>
   `${getBadgeCopySubpath(badgeType)}.description`
 
 const BadgeComponent: React.FC<Props> = ({ badge }) => {
-  const { t } = useTranslation('common')
+  const { t } = useTranslation('badge')
```
Lanny, line 184: You'll want to add 'badge' to the `namespacesRequired` array on the page component so these translations get included in the initial page load

packages/web/resolvers/thanks.ts
robin-macpherson: Ah yes! Thanks mate!

packages/web/components/Badge/Badge.tsx
robin-macpherson: Ah yes, I meant to come back and do this, what do you think? I made it so that the alt text should be "<badgeTitle>: <badgeDescription>"

packages/web/components/Badge/Badge.tsx
robin-macpherson: Great catch, thank ya!

packages/web/resolvers/thanks.ts
robin-macpherson: (Wrote offline but adding here too) Great call! I'm thinking we can do this as a fast-followup immediately after, along with a couple of other things, WDYT?

PR #772, opened 2022-02-01
Fix autosave

### LINKED ISSUE #755: 🐞 Auto-saving Broken In The Journaly Editor 🚨
#### Description

Auto-saving is broken in the Journaly Editor. I also think it should be expanded to include topics.

#### Repro Steps

1. Start typing some text into the body of a new post
2. Refresh the page - EXPECTED BEHAVIOR = the text you had written should be restored
3. Observe that the text is completely gone

Lanny:
Eliminate second render introduced by restoring `useAutosaveState` and fix autosave state by not even trying to SSR the editor.

## Description

Fixes #755

Either some change in slate or unfortunate interaction of slate-plugins/plate and slate caused slate to fail to update DOM state when a new `value` prop is passed to slate-react's `Slate` component, which broke autosaving (strictly speaking autosaving always worked fine, but slate just wouldn't reflect a restored value correctly, which you could observe with e.g. post titles saving and restoring correctly). To fix this we need to make sure that slate has the autosaved value available on its very first render. Further the DOM state of the editor must be equal during SSR and on the first render in the client or else slate explodes. To make this work, given save state is not available on the server, the best solution I could come up with is simply not to render the editor during SSR and let its first render on the client have the restored state. This seems like a least-of-several-evils solution.

## Subtasks

- [ ] ...

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

No migs

### DIFF
--- a/packages/web/components/JournalyEditor/JournalyEditor.tsx
+++ b/packages/web/components/JournalyEditor/JournalyEditor.tsx
@@ -68,6 +68,10 @@ const JournalyEditor = ({
     ;(slateRef as React.MutableRefObject<Editor>).current = editor
   }, [editor])
 
+  if (typeof window === 'undefined') {
+    return null
+  }
+
   return (
     <div className="editor-wrapper">
       <div className="editor-container">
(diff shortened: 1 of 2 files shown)


robin-macpherson (approved):
Woohooooo awesome thanks man! I was having a hard time figuring this one out. I guess this is one to keep an eye on as we continue to follow the Slate updates

PR #780, opened 2022-04-11
Make wonky notification data not cause the universe to implode.

Lanny:
No migs, just a bandaid to make wonk'd notifications not cause the whole page to crash


robin-macpherson (approved):
Epic, thanks mate!

Issue #785, opened 2022-05-06
📐 Modal Design Improvements


robin-macpherson:
### Description

I think our modal could have some design improvements – the biggest one being that having our usual header typography looks totally out of place. I think at a minimum we should update this to use our sans-serif font and make it a little smaller so this looks nice and sharp finally!

PR #786, opened 2022-05-08
📐 Modal Design Improvements ✐

robin-macpherson:
## Description

**Issue:**
- #785 

Our `<Modal>` component was built quite quickly very early on when we only really had one use case (gathering a user's language info when first signing up), but especially as we've expanded its use throughout the app, I've always felt like it needed a nice little design makeover. In particular the typography and spacing didn't feel right, so this PR makes the following improvements:
- Switches the font-family over to our sans-serif typeface
- Improves the spacing by removing the unnecessary huge amount of padding at the top when the header title isn't being shown
- Also reduces padding at the sides and bottom of the modal body to make things feel more balanced and spacious
- Centers title text

I think even this very simple change makes a big difference and the modal feels so much nicer and more spacious now without all that space being taken up by unnecessary padding!

## Subtasks

- [x] Adjust typography
- [x] Adjust spacing/layout
- [x] Test our various use cases

### Deployment Checklist

- [x] 🚨 Deploy code to stage
- [ ] 🚨 Deploy code to prod

## Migrations

N/A

## Screenshots

**1. Before**

**1. After**

**2. Before**

**2. After**
image

**3. Before**

**3. After**

### DIFF
--- a/packages/web/components/Modal/ModalBody.tsx
+++ b/packages/web/components/Modal/ModalBody.tsx
@@ -38,7 +38,7 @@ const ModalBody: React.FC<Props> = (props) => {
         }
         @media (min-width: ${modalConstants.modalBreakpoint}) {
           #modal-body {
-            padding: 0 64px 64px;
+            padding: 0 40px 40px;
           }
         }
 
--- a/packages/web/components/Modal/ModalBody.tsx
+++ b/packages/web/components/Modal/ModalBody.tsx
@@ -44,7 +44,10 @@ const ModalBody: React.FC<Props> = (props) => {
 
         h1 {
           margin-bottom: 16px;
-          ${theme.typography.headingXL};
+          ${theme.typography.fontFamilySansSerif};
+          font-size: 24px;
+          text-align: center;
+          font-weight: 600;
         }
       `}</style>
     </div>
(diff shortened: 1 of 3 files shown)


robin-macpherson:
Hey @Lanny ! Would love to get your thoughts on this. Very minor code changes but in my opinion it really makes a difference improving the modal. These improvements will be particularly important with #782 and related upcoming work where the modals will be used in more functional ways rather than purely informational.

packages/web/components/Modal/ModalBody.tsx
```diff
@@ -38,13 +38,16 @@ const ModalBody: React.FC<Props> = (props) => {
...
-            padding: 0 64px 64px;
+            padding: 0 40px 40px;
           }
         }
 
         h1 {
           margin-bottom: 16px;
-          ${theme.typography.headingXL};
+          ${theme.typography.fontFamilySansSerif};
+          font-size: 24px;
```
robin-macpherson, line 48: Can we use a variable here?

Lanny (approved):
Looks good! I like the sans for the title, kinda felt like I was reading a medieval scroll or something when confirming I wanted to delete a comment haha

PR #788, opened 2022-05-17
Write Custom `useDebouncedCallback` Hook & Partially Remove Lodash Dependency

robin-macpherson:
## Description

- A quick POC to look into simply creating our own custom hook for debouncing to partially remove our `lodash` dependency
- Removes a couple of imports where we were brining in all of `lodash` instead of importing a specific function

## Subtasks

- [x] Write the hook
- [x] Test the hook in My Feed with `<SearchInput>`
- [ ] Test the hook with `visualViewportDiff` in `<Toolbar>`
- [ ] A bit of TypeScript massaging with the hook

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

### DIFF
--- a/packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
+++ b/packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
@@ -31,6 +31,7 @@ import { options, isTableActive } from '../helpers'
 import SwitchToggle from '@/components/SwitchToggle'
 import { useTranslation } from '@/config/i18n'
 import useIntersectionObserver from '@/hooks/userIntersectionObserver'
+import useDebouncedCallback from '@/hooks/useDebouncedCallback'
 
 type ToolbarProps = {
   allowInlineImages: boolean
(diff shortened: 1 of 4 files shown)


packages/web/hooks/useDebouncedCallback.ts
```diff
@@ -0,0 +1,31 @@
...
+    (...args) => {
+      const later = () => {
+        clearTimeout(timeout.current)
+        callback(...args)
+      }
+
+      clearTimeout(timeout.current)
+      timeout.current = setTimeout(later, wait)
+    },
+    [callback, wait],
```
Lanny, line 27: because `callback` here is in the dependency array, to achieve a stable value the caller will have to memoize their function (likely using `useCallback`) before passing it to this function, so we don't achieve much by the `useCallback` call.

Since callers will have to memoize anyway (and the `useRef` here can be accomplished with just an object declared outside the returned function) I'd suggest making this simply be a function rather than a hook that callers can use in a `useMemo`.

packages/web/components/JournalyEditor/Toolbar/Toolbar.tsx
```diff
@@ -65,14 +66,16 @@ const Toolbar = ({
       setViewportsDiff(visualViewport.offsetTop)
     }
 
-    onVisualViewportChange()
+    const debouncedOnVisualViewportChange = useDebouncedCallback(onVisualViewportChange, 500)
```
Lanny, line 69: Calling a hook inside a `useEffect` callback is invalid, which contributes to the case for this being a simple function rather than a hook.

packages/web/hooks/useDebouncedCallback.ts
```diff
@@ -0,0 +1,31 @@
...
+  const timeout = useRef()
+
+  return useCallback(
+    (...args) => {
+      const later = () => {
+        clearTimeout(timeout.current)
+        callback(...args)
+      }
+
+      clearTimeout(timeout.current)
```
Lanny, line 24: So the behavior that's implemented here is that any call to the debounced function will wait at least `wait` ms before invoking the wrapped function, and no invocation will happen if another call to the debounced function happens before invocation of the wrapped func.

This is apparently lodash's definition of "debounce", but I think the behavior we actually want is what lodash calls "throttle", meaning the function will be invoked at most once in a period of `wait` ms, ensuring the last call to the throttled function invoke the wrapped function as soon as `wait` ms has elapsed since the last invocation.

So the logic for that would be something like:
- check the time since the last invocation of the wrapped function
- If it's more than `wait` ms ago, invoke again and update the last-invoked time
- If it's less than `wait` ms ago:
  - Cancel any pending calls (`clearTimeout()`)
  - Create a new pending call scheduled for `wait - (now - lastInvocation)` ms in the future
    - When that pending call does eventually execute, update the last-invoked time
 
This is based on the assumption that in the scrolling case, we do want to continue to call this function while scroll events are happening and continue updating the toolbar position, not wait until all the scrolling action has stopped, we just want to limit the frequency at which that happens

Issue #711, opened 2021-12-09
Don't Allow Empty Comments


robin-macpherson:
#### Description

Currently, it is possible to create empty comments (both Feedback & Post Comments) by simply clicking the submit button.

#### Proposed Solution

Let's do a simple check to make sure the comment body isn't empty and not allow empty comments

Lanny:
Partially fixed (for the feedback comment case) by #728

PR #790, opened 2022-06-12
#711 Don't Allow Empty Comments

Demezis:
## Description

**Issue:** #711 Don't Allow Empty Comments

## Subtasks

- [x] I have added validation to post comments
- [x] I have added validation to threads

### Deployment Checklist

- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy code to prod

## Screenshots
Animation

### DIFF
--- a/packages/web/components/Dashboard/Post/PostComment.tsx
+++ b/packages/web/components/Dashboard/Post/PostComment.tsx
@@ -48,6 +48,13 @@ const PostComment: React.FC<PostCommentProps> = ({
   })
 
   const updateExistingComment = () => {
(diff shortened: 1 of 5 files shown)


robin-macpherson (changes requested):
Thanks a lot for this PR @Demezis , it's great to see you contributing again! We're using a `ref` to track the comment body state for performance reasons and I think the refactor there is not worth the trade-off. To be honest I think for this we can just return early/do nothing if someone attempts to submit a comment and the ref value is empty.

Demezis:
I have reverted changes regarding threads, and now it should be fine.

PR #829, opened 2023-03-09
Add new Jadmin script for deleting a post

### LINKED ISSUE #828: 🧹 Write Script For Deleting A Post ONLY & Its Relational Data 💻
#### Problem

- We recently wrote a script for removing spam users and all of their related data, but there may be times we need to only delete a post and not the user.

#### Solution

- Let's write a script we can run that can handle all of this for us.

robin-macpherson:
## Description

**Issue:** 
* closes #828 

In #824 we added a script that deletes a user (intended for those who brazenly violate our Terms of Service) and all related records. There are some cases where we may want to delete only a post, so this PR adds a script to handle that.

## Subtasks

- [x] Add the new script
- [x] Test on local data

### Deployment Checklist

- [x] ~🚨 Publish new j-db-client version and update in both `web` and `j-mail`~
- [x] ~🚨 Deploy migs to stage~
- [ ] 🚨 Deploy code to stage
- [x] ~🚨 DEPLOY `j-mail` to stage~
- [x] ~🚨 Deploy migs to prod~
- [ ] 🚨 Deploy code to prod
- [x] ~🚨 DEPLOY `j-mail` TO PROD~

## Migrations

N/A

## Screenshots
N/A

### DIFF
--- a/packages/jadmin/jadmin.js
+++ b/packages/jadmin/jadmin.js
@@ -239,11 +239,6 @@ yargs
       await db.pool.end()
     },
   })
-  .demandCommand(1)
-  .parse(process.argv.slice(2))
-
-  yargs
-  .scriptName('jadmin')
   .command({
     command: 'delete-post <postId>',
     describe: 'Hard delete a post and related records',
--- a/packages/jadmin/jadmin.js
+++ b/packages/jadmin/jadmin.js
@@ -241,3 +241,93 @@ yargs
   })
   .demandCommand(1)
   .parse(process.argv.slice(2))
+
+  yargs
+  .scriptName('jadmin')
+  .command({
+    command: 'delete-post <postId>',
+    describe: 'Hard delete a post and related records',
+    builder: (yargs) => {
+      yargs.option('force', {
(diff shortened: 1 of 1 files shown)


packages/jadmin/jadmin.js
```diff
@@ -241,3 +241,93 @@ yargs
   })
   .demandCommand(1)
   .parse(process.argv.slice(2))
+
+  yargs
+  .scriptName('jadmin')
+  .command({
```
Lanny, line 247: this `.command` call should be chained on after the previous one, rather than have a whole additional `yags...parse(...)` call. So like: 

```
yargs
  .scriptName('jadmin')
  .command({
    command: 'delete-user <userId>',
    ...
  })
  .command({
    command: 'delete-post <postId>',
    ...
  })
  .demandCommand(1)
  .parse(process.argv.slice(2))
```

robin-macpherson: Awesome thanks! How's this?

Lanny (approved):

PR #832, opened 2023-03-23
Create Script For Awarding A Badge To A User

robin-macpherson:
## Description

Adds a Jadmin script for assigning badges to users!

## Subtasks

- [x] Add the new script
- [x] Test, fail, fix!
- [x] Test in prod
- [x] Fail and fix again!

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

No miggies ☺️

## Screenshots

N/A

### DIFF
--- a/packages/jadmin/jadmin.js
+++ b/packages/jadmin/jadmin.js
@@ -330,5 +330,78 @@ yargs
       await db.pool.end()
     },
   })
+  .command({
+    command: 'assign-badge <userId> <badgeName>',
+    describe: 'Assign a badge to a user',
+    builder: (yargs) => {
+      yargs.option('force', {
+        describe: 'Bypass interactive safety checks',
+        type: 'boolean',
+      })
+      yargs.option('database-url', {
+        describe: 'The URL to connect to the database',
+        type: 'string',
+      })
+    },
+    handler: async (args) => {
+      const dbUrl = args.databaseUrl || process.env.DATABASE_URL
+
+      if (!dbUrl) {
+        console.error('No database URL has been provided')
+        process.exit(1)
+      }
+
+      const parsedDbUrl = parse(dbUrl)
+      let promptResponse
+
+      if (!['localhost', '127.0.0.1'].includes(parsedDbUrl.host) && !args.force) {
+        promptResponse = await prompts({
+          type: 'confirm',
+          name: 'confirmDbUrl',
+          message: `CAUTION: your $DATABASE_URL is currently set to a remote database: ${dbUrl}. Are you sure you want to continue?`,
+        })
+
+        if (!promptResponse.confirmDbUrl) return
+      }
+
+      const db = pgTag(
+        new Pool({
+          connectionString: dbUrl,
+        }),
+      )
+
+      const userId = parseInt(args.userId)
+
+      const user = await db.get`
+        SELECT *
+        FROM "User"
+        WHERE id = ${userId}
+      `
+
+      if (!user) {
+        console.error(`No User with ID ${userId} found`)
+        process.exit(1)
+      }
+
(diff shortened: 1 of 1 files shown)


packages/jadmin/jadmin.js
```diff
@@ -330,5 +330,78 @@ yargs
...
+        })
+
+        if (!promptResponse.confirmDbUrl) return
+      }
+
+      const db = pgTag(
+        new Pool({
+          connectionString: dbUrl,
+        }),
+      )
```
Lanny, line 371: We do this DB check in all three scripts now I think, and I think it's exactly the same check, so this would be great to factor out:

```suggestion
    const async getDb = (args) => {
      const dbUrl = args.databaseUrl || process.env.DATABASE_URL

      if (!dbUrl) {
        console.error('No database URL has been provided')
        process.exit(1)
      }

      const parsedDbUrl = parse(dbUrl)
      let promptResponse

      if (!['localhost', '127.0.0.1'].includes(parsedDbUrl.host) && !args.force) {
        promptResponse = await prompts({
          type: 'confirm',
          name: 'confirmDbUrl',
          message: `CAUTION: your $DATABASE_URL is currently set to a remote database: ${dbUrl}. Are you sure you want to continue?`,
        })

        if (!promptResponse.confirmDbUrl) return
      }

      const db = pgTag(
        new Pool({
          connectionString: dbUrl,
        }),
      )
      
      return db
    }
    
    ... inside handlers ...
    
    const db = await getDb(args)
```

Lanny (approved):

packages/jadmin/jadmin.js
robin-macpherson: Great call! Done ✅✏️🦔

robin-macpherson:
@Lanny it's interesting, testing the script against the local DB everything works fine, but when I try it against prod I get a stack trace with the following in it:

`$ ./jadmin.js assign-badge 9 "SPAM_REPORTER"`
`> error: invalid input syntax for type integer: "SPAM_REPORTER"`

I've tried logging all the values at various stages... it almost seems like perhaps the syntax I'm using for those `DEFAULT` values or something isn't being interpreted in the same way for the prod DB? Any ideas?

packages/jadmin/jadmin.js
```diff
@@ -330,5 +308,58 @@ yargs
...
+      const badgeOptionsArray = badgeOptions.split(',')
+      console.log('badgeName', args.badgeName, typeof args.badgeName)
+      console.log('badgeOptionsResponse', badgeOptionsResponse)
+      console.log('badgeOptions', badgeOptions)
+      console.log('badgeOptionsArray', badgeOptionsArray)
+      if (badgeOptionsArray.includes(args.badgeName)) {
+        await db.query`
+          INSERT INTO "UserBadge"
+          VALUES (DEFAULT, ${args.badgeName}, DEFAULT, ${userId});
+        `
```
Lanny, line 354: Try this:

```suggestion
        await db.query`
          INSERT INTO "UserBadge"
          ("type", "userId")
          VALUES
          (${args.badgeName}, ${userId});
        `
```

When you do don't name the column names (`("type", "userId")`) then the assumption is that the VALUES sequence is provided in the order they were listed in the table's CREATE statement. In prod that seems to be createdAt, id, type, userId and based on this I think your local data got id, type, createdAt, userId. From prod:

```
journaly_db=> SELECT * FROM "UserBadge" LIMIT 1;
        createdAt        | id |    type    | userId 
-------------------------+----+------------+--------
 2020-11-09 21:00:22.936 |  1 | ALPHA_USER |     29
 ```

PR #840, opened 2023-04-08
⚙️ Build Initial User Configuration - Notification Settings

### LINKED ISSUE #527: ⚙️ Control over Daily Digest Email
This is a small change only, but I feel it would be better for users if they had the opportunity to change the frequency of the emails that arrive in their inboxes.

An email a day is quite high, so an option for emails every *N* days may be more useful.

robin-macpherson:
## Description

**Issue:**
- closes #527 

### Background

* One thing we haven't built as of yet is the ability for users to configure their preferences, the primary one being *Notification Preferences*, so that they can choose which notifications they receive, how often, etc.
* Ever since we launched, some users have asked for the ability to turn off the daily digest email or reduce it's frequency – although I should note we really haven't had too many of these requests.
* When we launched in-app notifications, we did have some users comment that now the daily digest email wasn't necessary for them.

### This PR

* Builds out the foundation for having user-level configuration with a new `UserConfiguration` table where we can store a variety of different settings in the future.
* Currently, there is one main column: `digestEmail` whose type is a new enum: `DigestEmailConfiguration`.
* Long-term, I think it would be nice to also offer a "Weekly Digest" email, so I have included a `WEEKLY` value in th enum, but I have hidden this option in the U.I.
* Builds out support for having the digest email either be `DAILY` or to turn it off.

## Subtasks

- [x] Technical/Database Design
**Backend:**
- [x] Add new table to schema and generate migration
- [x] Write backfill SQL and add to the migration
- [x] Add new values to GraphQL Schema
- [x] Test everything is working so far
**Frontend:**
- [x] Add new `NotificationSettingsForm` section to `Settings > Profile > Account`
- [x] Create new `DigestEmailConfigurationFormField` component and get the basics all working
- [x] Restrict options to `DAILY` & `OFF`
- [x] Add a mutation for updating configuration/settings
- [x] Add a toast confirmation/error message
- [x] Add loading spinner for account page
- [x] Consider cleaning up `currentUser` > `data.currentUser...` on profile page

### Deployment Checklist

- [x] 🚨 Deploy migs to stage
- [x] 🚨 Deploy code to stage
- [x] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

* Contains a migration to add a new table: `UserConfiguration`
* Also contains backfill SQL to add a record to that table for all users
* Default value will be `DAILY` for all users upon this migration & backfill

## Screenshots

**A - New Form with `DAILY` selected (default)
image

**B - New Form with `OFF` selected**
image


packages/web/components/DigestEmailConfigurationFormField/DigestEmailConfigurationFormField.tsx
```diff
@@ -0,0 +1,82 @@
...
+  // until we build support for it
+  const options = (Object.keys(DigestEmailConfigurationType) as (keyof typeof DigestEmailConfigurationType)[]).map(
+    (key, index) => ({
+      value: index,
+      displayName: DigestEmailConfigurationType[key],
+    }),
+  ).filter((option) => option.displayName !== DigestEmailConfigurationType.Weekly);
+
+  // TODO: Figure out why this is -1 even though loggig everything before the OR returns 0
+  const currentConfig = options.find((option) => option.displayName === digestEmailConfig)?.value || -1
```
Lanny, line 28:
```suggestion
  const currentConfig = options.find((option) => option.displayName === digestEmailConfig)?.value ?? -1
```

Since `0` is falsy `0 || -1` will evaluate to `-1`.  So using the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) should fix this

packages/web/components/DigestEmailConfigurationFormField/DigestEmailConfigurationFormField.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [selectedOptionId, setSelectedOptionId] = useState<number>(currentConfig)
+  console.log(digestEmailConfig)
+  const [
+    addUserInterest,
+    { loading: addingUserInterest },
+  ] = useAddUserInterestMutation()
```
Lanny, line 34: We probably want to change the pattern from what's being copied here so that `NotificationSettingsForm` owns the state and is responsible for patching the settings record. Not sure if you want "commit on change" semantics or for there to be an explicit save CTA here, but either way having a mutation and logic-heavy component per-setting seems like something we'd want to avoid.

packages/web/graphql/fragments.graphql
```diff
@@ -9,6 +9,9 @@ fragment UserFragment on User {
   city
   country
   emailAddressVerified
+  configuration {
+    ...UserConfigurationFragment
+  }
```
Lanny, line 14: We user this fragment in several places, including profile pages. We probably don't want to to include the general case. If we don't already have one, maybe a "SettingsPageUserFragment" would make sense? Or maybe having this only be a part of the top level query instead of including it in any fragment?

packages/web/resolvers/user.ts
```diff
@@ -63,6 +63,7 @@ const User = objectType({
     t.model.createdAt()
     t.model.membershipSubscription()
     t.model.isStudent()
+    t.model.configuration()
```
Lanny, line 66: This will expose this field publicly for everyone. I know giving up "what is this person's email digest preference" isn't exactly sensitive info, but it just doesn't quite feel right to make settings public record. You could use a similar pattern to email to do an auth check here.

packages/web/resolvers/utils/notifications.ts
```diff
@@ -18,7 +19,10 @@ const createEmailNotification = (
   db: PrismaClient,
   user: User,
   note: EmailNotificationCreationType,
+  digestEmailConfiguration: DigestEmailConfiguration,
 ) => {
+  if (digestEmailConfiguration === DigestEmailConfiguration.OFF) return
```
Lanny, line 24: I kinda think rather than avoiding creating the (email) notification record based on this, we should suppress sending the email based on the config preference. It's kind of a corner case, but if you turn off email notifications an hour before the email goes out, you'd expect the email not to be sent vs. this approach makes it so that you'd still get your digest email, it just won't contain the last hour of activity. The approach of collecting the notification records but not sending an email will also serve us for the weekly digest case where we still want to collect the records but only send the digest at the end of the week.

packages/web/components/DigestEmailConfigurationFormField/DigestEmailConfigurationFormField.tsx
robin-macpherson: Ahhh yes of course, thanks!

packages/web/resolvers/user.ts
robin-macpherson: Done!

packages/web/components/DigestEmailConfigurationFormField/DigestEmailConfigurationFormField.tsx
robin-macpherson: Good call!

packages/web/graphql/fragments.graphql
robin-macpherson: Yep good call too, moved to the settings data query thing!

packages/web/public/static/locales/de/settings.json
```diff
@@ -101,6 +101,14 @@
       "passwordChangeSuccess": "Das Passwort wurde erfolgreich geändert.",
```
robin-macpherson, line 101: Hi @LindasLinguas ! Do you think you could translate these new notification settings strings for us? 😀 Vielen Dank!

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
       "passwordChangeSuccess": "密碼更改成功",
       "updateError": "密碼更新失敗",
       "newPasswordCannotBeTheSameError": "新密碼不可與舊密碼相同"
+    },
```
robin-macpherson, line 110: @pushinhuang would you mind translating these new translation strings into `zh_TW` and also the `zh_CN` file as well? 謝謝～～～

packages/web/resolvers/utils/notifications.ts
robin-macpherson: Done!

packages/web/graphql/userConfiguration/updateUserConfiguration.graphql
```diff
@@ -0,0 +1,5 @@
+mutation updateUserConfiguration($digestEmailConfig: DigestEmailConfiguration) {
```
robin-macpherson, line 1: @Lanny just FYI I found that you actually can reference a custom type within the args in these definitions!

packages/web/resolvers/user.ts
```diff
@@ -419,6 +426,48 @@ const UserMutations = extendType({
       },
     })
 
+    t.field('updateUserConfiguration', {
+      type: 'UserConfiguration',
+      args: {
+        digestEmailConfig: arg({
```
robin-macpherson, line 432: @Lanny I also found that it was very straightforward here to do the same, simply using `arg()` and passing the string name of the enum for the `type` value. Simpler than we thought!

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
...
+    "notificationSettings": {
+      "legend": "Notifications",
```
pushinhuang, line 112:
```suggestion
      "legend": "通知",
```

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
```
pushinhuang, line 113:
```suggestion
      "digestEmailSelectLabel": "Email通知",
```

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
```
pushinhuang, line 114:
```suggestion
      "digestEmailSelectPlaceholder": "你希望收到Email通知的頻率?",
```

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "updateSettingsButtonText": "Update",
```
pushinhuang, line 115:
```suggestion
      "updateSettingsButtonText": "更新",
```

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
```
pushinhuang, line 116:
```suggestion
      "updateDigestEmailSuccessMessage": "你的設定已更新成功!",
```

packages/web/public/static/locales/zh_CN/settings.json
```diff
@@ -107,6 +107,14 @@
       "passwordChangeSuccess": "密码更改成功",
       "updateError": "密码更新失败",
       "newPasswordCannotBeTheSameError": "新密码不可与旧密码相同"
+    },
+    "notificationSettings": {
+      "legend": "Notifications",
```
pushinhuang, line 112:
```suggestion
      "legend": "通知",
```

packages/web/public/static/locales/zh_CN/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
```
pushinhuang, line 113:
```suggestion
      "digestEmailSelectLabel": "Email通知",
```

packages/web/public/static/locales/zh_CN/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
+      "updateSettingsButtonText": "Update",
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
```
pushinhuang, line 116:
```suggestion
      "updateDigestEmailSuccessMessage": "你的设定已更新成功!",
```

packages/web/public/static/locales/zh_CN/settings.json
```diff
@@ -107,6 +107,14 @@
       "passwordChangeSuccess": "密码更改成功",
       "updateError": "密码更新失败",
       "newPasswordCannotBeTheSameError": "新密码不可与旧密码相同"
+    },
+    "notificationSettings": {
+      "legend": "Notifications",
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
+      "updateSettingsButtonText": "Update",
```
pushinhuang, line 115:
```suggestion
      "updateSettingsButtonText": "更新",
```

packages/web/public/static/locales/en/settings.json
```diff
@@ -108,6 +108,14 @@
...
       "updateError": "There was an error updating your password.",
       "newPasswordCannotBeTheSameError": "new password cannot be the same as old"
+    },
+    "notificationSettings": {
+      "legend": "Notifications",
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
+      "updateSettingsButtonText": "Update",
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
+      "updateDigestEmailErrorMessage": "Your settings have been updated successfully!"
```
pushinhuang, line 118:
```suggestion
      "updateDigestEmailErrorMessage": "There was an error updating your settings."
```

packages/web/public/static/locales/de/settings.json
```diff
@@ -101,6 +101,14 @@
...
       "updateError": "Beim Ändern des Passworts ist ein Fehler aufgetreten.",
       "newPasswordCannotBeTheSameError": "Das neue Passwort darf nicht dasselbe wie das alte sein."
+    },
+    "notificationSettings": {
+      "legend": "Notifications",
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
+      "updateSettingsButtonText": "Update",
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
+      "updateDigestEmailErrorMessage": "Your settings have been updated successfully!"
```
pushinhuang, line 111:
```suggestion
      "updateDigestEmailErrorMessage": "There was an error updating your settings."
```

packages/web/public/static/locales/es/settings.json
```diff
@@ -82,6 +82,14 @@
...
       "newPasswordPlaceholder": "Nueva contraseña",
       "newPasswordError": "Escriba su nueva contraseña"
+    },
+    "notificationSettings": {
+      "legend": "Notifications",
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
+      "updateSettingsButtonText": "Update",
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
+      "updateDigestEmailErrorMessage": "Your settings have been updated successfully!"
```
pushinhuang, line 92:
```suggestion
      "updateDigestEmailErrorMessage": "There was an error updating your settings."
```

packages/web/public/static/locales/it/settings.json
```diff
@@ -106,6 +106,14 @@
...
       "updateError": "Si è verificato un errore nell'aggiornamento della tua password.",
       "newPasswordCannotBeTheSameError": "la nuova password non può essere uguale alla vecchia"
+    },
+    "notificationSettings": {
+      "legend": "Notifications",
+      "digestEmailSelectLabel": "Digest Email (Notifies you of all activity)",
+      "digestEmailSelectPlaceholder": "How often would you like to receive the digest email?",
+      "updateSettingsButtonText": "Update",
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
+      "updateDigestEmailErrorMessage": "Your settings have been updated successfully!"
```
pushinhuang, line 116:
```suggestion
      "updateDigestEmailErrorMessage": "There was an error updating your settings."
```

packages/web/public/static/locales/zh_TW/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "updateDigestEmailErrorMessage": "Your settings have been updated successfully!"
```
pushinhuang, line 117:
```suggestion
      "updateDigestEmailErrorMessage": "更新失敗"
```

packages/web/public/static/locales/zh_CN/settings.json
```diff
@@ -107,6 +107,14 @@
...
+      "updateDigestEmailSuccessMessage": "Your settings have been updated successfully!",
+      "updateDigestEmailErrorMessage": "Your settings have been updated successfully!"
```
pushinhuang, line 117:
```suggestion
      "updateDigestEmailErrorMessage": "更新失败"
```

packages/j-mail/src/handler.ts
```diff
@@ -33,10 +35,21 @@ const getDataForUpdateEmail = async (
   let thanksCount = 0
   let clapCount = 0
 
-  const user = (await prisma.user.findUnique({ where: { id: userId } })) as User
+  const user = (await prisma.user.findUnique({
+    where: {
+      id: userId,
+    },
+  })) as User
```
Lanny, line 42: I think you can do this:

```suggestion
  const user = await prisma.user.findUnique({
    where: {
      id: userId,
    },
    include: { configuration: true }
  })
```

and fetch the configuration object at the same time (saving a round trip to the DB)

packages/j-mail/src/handler.ts
```diff
@@ -121,20 +134,26 @@ export const sendUpdateEmails: Handler = async (event, context) => {
...
-      from: 'robin@journaly.com',
-      to: data.user.email,
-      subject: 'New Activity on Journaly 📖',
-      html: body,
-    })
+    // Currently we only support 'DAILY' OR 'OFF' so we simply
+    // skip enqueueEmail if 'OFF'.
+    // When we build support for 'WEEKLY' we'll need to make this
+    // logic a little more sophisticated.
+    if (data.userConfig.digestEmail !== DigestEmailConfiguration.OFF) {
```
Lanny, line 141: isn't `userConfig` nullable? Might need to optional chain here.

Since we are changing jmail (which is good!) we should probably deploy this to stage and confirm it works.

Lanny (approved):
Looks good man, left two comments, first is minor, second should be an easy fix. Do recommend trying it out on stage. You should be able to use the sls `invoke` command to make the email send job fire without having to wait for the cron job: https://www.serverless.com/framework/docs/providers/aws/cli-reference/invoke

Issue #808, opened 2022-12-23
🔐 Handle All Open Security Advisories


robin-macpherson:
#### Description

- We currently have a collection of open security advisories
- Let's close out all of these

PR #841, opened 2023-04-11
Replace slate-plugins with its successor, plate

Lanny:
## Description

**Issue:** #808

## Subtasks

- [ ] ...

### Deployment Checklist

- [x] ~🚨 Publish new j-db-client version and update in both `web` and `j-mail`~
- [x] ~🚨 Deploy migs to stage~
- [ ] 🚨 Deploy code to stage
- [x] ~🚨 DEPLOY `j-mail` to stage~
- [x] ~🚨 Deploy migs to prod~
- [ ] 🚨 Deploy code to prod
- [x] ~🚨 DEPLOY `j-mail` TO PROD~

## Migrations

N/A

## Screenshots

N/A


packages/web/components/JournalyEditor/JournalyEditor.tsx
```diff
@@ -105,6 +127,56 @@ const JournalyEditor = ({
...
+        }
+      `}</style>
+    </div>
+  )
+}
+
+export default JournalyEditor
+
+
+/*
```
robin-macpherson, line 179: 🧹 Cleanup Note 🧹

packages/web/components/JournalyEditor/RenderElement.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
-import { renderElementTable } from '@udecode/slate-plugins'
+//import { renderElementTable } from '@udecode/plate'
```
robin-macpherson, line 2: 🧹 Cleanup Note 🧹

robin-macpherson (commented):
Woohooo awesome man, thanks for this! For now I just did a super quick pass and left a couple of cleanup notes to help point out some commented out stuff to clean up, but will take a full pass tomorrow.

Also just checking did you run `nvm use` from within our project before regenerating the lock file? I know we have a lot of upgrades so might be expected but the delta looks pretty huge and I had noticed that when I took over the badge PR from you the lock file had a large diff but when I regenerated it myself after nvm use most of that went away, so just checking

robin-macpherson:
Hmm interesting, the redeployment failed, I suspected it was potentially just some env problems but staging was working fine, so I merged your code into staging and now the same situation there. I'll try checking your branch out locally tomorrow and see what's happening, I assume that was working fine for you

robin-macpherson:
Eh! Decided to give it a quick try before heading off to bed 😆

Here are all my findings:
A few findings so far:
- Some of the i18n translations seem to be broken, they are displaying the translation JSON key and not the value:
    - “My Posts” page: “publishedTab, privateTab, draftsTab”
    - “New Post” page: “typewriterSounds”
- In the toolbar, bold, italics, underline, and title all seem to work great. The quote text one seems to be broken, it doesn’t select properly when I click it and then something strange happens when I try to type after.
- Wow, noticed a weird bug but it’s in production too. Weird things happen when you try to turn on and especially turn off the B U I buttons. If you have B turned on and then click on I with your mouse… strange
- Hmm never mind, quote button seems to be working now?
- Table seems to work! Wow pretty cool little upgrade being able to choose the border options! I do notice that now the table doesn’t resize automatically.
- Tested French including accented character composition as well as the pre-composed accented characters. ✅
- Tested both Chinese and Japanese ✅
- Hyperlinks work
- Inline images work
- ❌ typewriter sounds aren’t working
- Oh! Was just about to wrap up and saw a big console error: `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`
- Agh, can’t manage to repro!

robin-macpherson:
WOOHOO! Very exciting testing results:
- ✅ IME bug on Firefox **FIXED**! 🎈 
- ✅ Typewriter sounds working now

One last unfortunate bug I found:
- ⭕️ While translations appear to be loading fine in Chrome now, they do not in FireFox (See screenshots)

Screenshot 2023-05-02 at 6 20 52 PM
Screenshot 2023-05-02 at 6 22 35 PM

packages/web/package.json
```diff
@@ -114,30 +115,34 @@
...
-    "eslint-plugin-prettier": "^3.1.3",
-    "eslint-plugin-react": "^7.24.0",
     "express": "^4.17.1",
     "husky": "^4.2.3",
     "jsonc-parser": "^3.0.0",
     "p-retry": "^4.2.0",
     "prettier": "^2.0.1",
     "pretty-quick": "^2.0.1",
-    "prisma": "3.5.0",
+    "prisma": "3.15.2",
```
robin-macpherson, line 132: What's the reason for the downgrade? I think we had upgraded to a certain version specifically to get Cascading delete functionality but I haven't gone back and checked exactly what that version was, have you double-checked we don't lose anything we need with the downgrade?

Lanny: The new version is 3.15.2, while the old one was 3.5.0, so that's an upgrade of 10 minor versions right? I think nexus needed something from a newer prisma/prisma-client version, hence the upgrade

robin-macpherson: Oh LOL bloody hell that was dumb sleepy Robin reviewing before bed 🤣 somehow I saw 3.1.5 😂

robin-macpherson (approved):
🪠⚙️🔧🦔

PR #857, opened 2023-05-11
🧹 Remove Unused Dependencies 🤖

### LINKED ISSUE #856: 🧹 Clean up unused dependencies
#### Description

While taking a more detailed look through some of our Dependabot alerts, I noticed one was based on a package we have called `Husky` which I ***seem*** to remember being something that was used way way back when we were using Prisma's framework that's now in a grave yard 🪦💀

But I wasn't 100% sure if I can confidently axe it, despite not seeing its usage anywhere, so I looked around and found the `depcheck` package to check for unused dependencies. Running it actually produced:

```
Unused dependencies
* @apollo/react-ssr
* @types/cors
* @types/nodemailer
* date-fns-tz
* eslint-config-next
* tsconfig-paths
Unused devDependencies
* @graphql-codegen/typescript
* @graphql-codegen/typescript-operations
* @graphql-codegen/typescript-react-apollo
* @svgr/plugin-jsx
* @svgr/plugin-prettier
* @svgr/plugin-svgo
* colors
* husky
* p-retry
* pretty-quick
Missing dependencies
* nodegit: ./tt.js
* react-i18next: ./hooks/useUILanguage.ts
* graphql-tag: ./generated/graphql.tsx
* csstype: ./components/Popover/Popover.tsx
```

I think we should go through this list and look into cleaning up/removing unused dependencies.

robin-macpherson:
## Description

**Issue:**
* closes #856 

See the linked issue for full details – I started poking around our dependencies and found that there appear to be a pretty large number that are no longer being used. In some cases I think this is due to code changes where we stopped using them, but when I looked into a few of them in more detail, the PRs where they were added don't seem to actually introduce any usage of them, unless I'm misunderstanding something.

I definitely want to sanity check myself here with a thorough review, but from what I can see these aren't being used anywhere, the build still passes and the app appears to run fine.

I have to say though some of these were a bit surprising (`@apollo/react-ssr`??) so I'm curious to see what you think @Lanny !

## Subtasks

- [x] Delete first round and test the build/app
- [x] Delete second round and test

### Deployment Checklist

- [x] 🚨 Deploy code to stage
- [x] 🚨 Deploy code to prod

## Migrations

N/A

## Screenshots

N/A


robin-macpherson:
@Lanny UPDATE: as usual with NPM stuff 😑... I was kind of surprised that the lock file changes weren't more substantial, so tried blowing away my `node_modules` and regenerating the lock file, and sure enough the changes were way larger and **now** I do actually get some failures when running the build:

```
./components/GenericErrorBoundary/GenericErrorBoundary.tsx:37:17
Type error: Type 'T' is not assignable to type 'IntrinsicAttributes & T'.
  Type 'T' is not assignable to type 'IntrinsicAttributes'.

  35 |       if (this.props.FallbackComponent) {
  36 |         const { FallbackComponent, fallbackProps } = this.props
> 37 |         return <FallbackComponent {...fallbackProps} />
     |                 ^
  38 |       }
  39 |
  40 |       return (
```

PR #871, opened 2023-09-08
A smorgasbord of a11y changes to nav and the new post page

Lanny:
## Description

point-by-point re: the email that spawned this effort:

>When I try to add a photo via Unsplash I don't seem to be able to access the tab. In the screencast attached, I'm tabbing repeatedly at a consistent speed but only the 'choose image to upload' and 'close' links are highlighted. After a bit of trial and error I think at one point I *can* hit return to access the 'search Unsplash' option but if the active element was always highlighted, this would be a piece of cake and one less barrier :)

Fixed, although the approach was to use an underline (same as the hover state) to signal focus on these links. I think the looks reasonable but maybe it's not super duper obvious. Open to considering a more visually impactful signal of focus if this doesn't work

>Similarly, when I'm on the New Post page, I can tab through the fields (title, language, etc) much more logically, but the two dropdowns aren't highlighted when active either. I can use access them if I'm paying close attention and use alt-down to open the list and navigate up and down them.

Fixed issue with focus not being visually indicated, looks like this:

image

Only change here was to remove our override of the browser default focus indicator (so behavior is a little different between browsers, e.g. in chromium the focus indicator is orange :shrug: ). I think it looks fine, but then I also am a semi-regular tab-er so I generally value an indication of focus position over aesthetics

WRT the keyboard interaction for actually selecting a value, the language select seems wholly reasonable to me. The multi-select ones (i.e. topics) are kinda janky, if you focus and then use the arrow keys, each keypress results in a new topic being added. On FF on linux using alt+down actually behaved how I'd expect, but we should probably fix the "press arrow while focused" case, which I didn't do, will need to think about what the appropriate behavior here is and what's within reach to do.

>However, when I try to escape out of the list (either with Escape or alt-up or tab) the tag I'm on is automatically added. I haven't yet found a way to close the list without adding another tag, though maybe there is one?! It means I have to close the list then go and delete the last random tag added (I'm usually still browsing to see if there's another appropriate tag) before carrying on.

Specifically around exiting, this is all browser controlled. Linux/FF allowed me to alt+down, move through options without selecting anything, and press escape to close the select without picking anything. I'm not sure we have a lot of control over this, although we could probably improve the multi-select experience if only by disabling raw arrow inputs while focused. 

>I also wonder if the username could *not* be an active link in this screen? It's a bit odd that the tab/element order takes you to the username before the 'choose image' element, and I don't see a benefit of hyperlinking the username as it can't be changed here.

This is somewhat intentional behavior. The username is in fact a link, so it should be focusable for comparable functionality to what mouse users get. re: it not being changeable, that's true, but we keep it as a link here because it will be a link in the final post and we're going for a "preview" kinda effect here. I'm going to make the case for keeping it focusable on the grounds that it's a better parallel to the mouse-using experience (and actually kinda important when looking at a published post) but I did change the tab order to be what you'd expect as a top-down RTL reader.

> Bypass blocks would be a fantastic thing to add in, too (see [2.4.1 bypass blocks](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=241#navigable)) as I do spend a lot of time going through a) the left-hand menu and b) the formatting bar.
The left-hand menu seems to have an error in the element labelling too, as tabbing skips Notifications, Log out, and Help (I've added a screencast of that too).

Added skip links to both the nav and the editor toolbar.

**Issue:** <#issue>

...

Journaly.com profile url (include if you'd like the "code contributor" badge):

## Subtasks

- [ ] ...

### Deployment Checklist

- [ ] 🚨 Deploy migs to stage
- [ ] 🚨 Deploy code to stage
- [ ] 🚨 Deploy migs to prod
- [ ] 🚨 Deploy code to prod

## Migrations

### DIFF
--- a/packages/web/components/Button/Button.tsx
+++ b/packages/web/components/Button/Button.tsx
@@ -36,6 +36,7 @@ type Props = {
   variant?: ButtonVariant
   style?: CSS.Properties | undefined
   dataTestId?: string
+  tabIndex?: number
 }
 
 const Button: React.FC<Props> = (props) => {
@@ -172,10 +173,9 @@ const Button: React.FC<Props> = (props) => {
           border: none;
           padding: 0;
         }
-        .${ButtonVariant.Link}:hover:not(:disabled) {
-          text-decoration: underline;
-        }
+        .${ButtonVariant.Link}:hover:not(:disabled),
         .${ButtonVariant.Link}:focus:not(:disabled) {
+          text-decoration: underline;
           outline: none;
           border-width: 0;
         }
--- a/packages/web/components/Dashboard/Nav/Nav.tsx
+++ b/packages/web/components/Dashboard/Nav/Nav.tsx
@@ -153,26 +153,26 @@ const Nav: React.FC<Props> = ({ expanded, collapse, disableLargeNav }) => {
               </div>
               <div className="nav-bottom">
                 <NavLink href="/my-feed">
-                  <a className="nav-link" onClick={handleCollapse} data-testid="my-feed-nav-link">
+                  <span className="nav-link" onClick={handleCollapse} data-testid="my-feed-nav-link">
                     <FeedIcon aria-hidden="true" />
                     <span className="nav-link-text">{t('dashboardNav.myFeed')}</span>
-                  </a>
+                  </span>
                 </NavLink>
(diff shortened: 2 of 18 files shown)


robin-macpherson (approved):
🦔🔥✨

